PR #2, opened 2023-05-09
Arrange bulkAlloc.h and Add Memory opt section to README.

moticless:


src/bulkAlloc.h
```diff
@@ -62,24 +62,20 @@
...
-/* debug */
 void serPoolPrintDbg(RdbParser *p);
 int serPoolIsNewNextAllocDbg(RdbParser *p);
 
-/* Unmanaged */
-void allocUnmanagedBulk(RdbParser *p, size_t len, AllocUnmngTypeRq rq, char *refBuf, BulkInfo *bi);
-void freeUnmanagedBulk(RdbParser *p, BulkInfo *binfo);
+/*** Unmanaged allocation ***/
+void unmngAllocBulk(RdbParser *p, size_t len, AllocUnmngTypeRq rq, char *refBuf, BulkInfo *bi);
+void unmngFreeBulk(RdbParser *p, BulkInfo *binfo);
```
oranagra, line 76: i preferred the previous one, "unmng" is not a thing.

oranagra (dismissed):

src/bulkAlloc.h
moticless: I was looking for a common prefix that is aligned with the header file description and API and makes clear distinction  between serialized-pool bulk management vs. unmanaged bulks ( `serPoolXXX` vs `unmngXXX`). 

For this case, I personally prefer clartiy over grammar correctness.

oranagra: so maybe they should all have a `bulkAlloc` prefix, and it should be `bulkAllocUnmanagedXXXX` and `bulkAllocSerializedPoolXXX`.
or something shorter but still using complete words or common terms.
i.e. `allocPoolXXXX` and `allocUnmanagedXXX` or alike.
Maybe "unmanaged" can be replaced with "user"?

oranagra: fine.. if it makes sense to you (i'm not following this closely at the moment, too busy).
if we can avoid the `unmng` abbreviation i'd prefer it. and if we can have a common prefix for both allocator that's also preferred.

oranagra (approved):
skimmed though the new API prefixes, LGTM

PR #5, opened 2023-06-11
Refine rdb-cli, add RESP reader and allow config pipeline-depth

moticless:
• Create basic RESP reader to parse responses from Redis server
• rebrand rdb-convert to rdb-cli. Arrange its usage to be more structured.
• Support configurable pipeline-depth when working with live Redis server.


src/ext/readerResp.c
```diff
@@ -0,0 +1,256 @@
...
+    UNUSED(buffInfo);
+    snprintf(ctx->errorMsg,sizeof(ctx->errorMsg),"Unexpected bulk reply");
+    return RESP_REPLY_ERR;
+}
+
+static RespRes readRespReplyAggregate(RespReaderCtx *ctx, RespReplyBuff *buffInfo) {
+    UNUSED(buffInfo);
+    snprintf(ctx->errorMsg,sizeof(ctx->errorMsg),"Unexpected aggregate reply");
+    return RESP_REPLY_ERR;
+}
```
oranagra, line 150: am i reviewing too early?

src/ext/respToTcpLoader.c
```diff
@@ -9,72 +9,125 @@
...
+        } else {
+            /* Error occurred or no data available */
+            if (errno == EWOULDBLOCK || errno == EAGAIN) {
+                /* No data available, retry after delay */
+                if (retries >= MAX_RETRIES) {
+                    RDB_reportError(ctx->p, (RdbRes) RDBX_ERR_RESP2TCP_MAX_RETRIES, "Maximum number of retries reached. Exit.");
+                    return 1;
+                }
+                usleep(delayUs);
+                delayUs *= 2;  /* Double the delay for each retry */
```
oranagra, line 83: maybe i'm out of sync with the recent conversation about this (i remember seeing something), or maybe i didn't read all the code yet, but we do plan to get rid of this sleep right?
i.e. we should allow some pipeline of requests until we read responses, but when we eventually decide to block on a response, it should be a blocking IO (not a sleep).

please point me to what i'm missing.

src/ext/respToTcpLoader.c
```diff
@@ -9,72 +9,125 @@
...
-#define NUM_PENDING_CMDS_LOW_THRESHOLD  100
+#define NUM_PENDING_CMDS_HIGH_THD 200
+#define NUM_PENDING_CMDS_LOW_THD  100
+
+#define NUM_RECORDED_CMDS         400
+#define RECORDED_MAX_SIZE         40
+#define REPLY_BUFF_SIZE           4096
+
+#define INITIAL_DELAY_USEC        20
+#define MAX_RETRIES               6
```
oranagra, line 28: better add some docs about all of these, what they do and what are the units.

src/ext/respToTcpLoader.c
```diff
@@ -9,72 +9,125 @@
...
     return 0;
 }
 
-/* return 0 for success. 1 Otherwise. */
-int tcpLoaderFlush(void *context) {
-    RdbxRespToTcpLoader *ctx = context;
-    if (ctx->pendingCmds.num)
-        return readReplies(ctx, ctx->pendingCmds.num);
-    return 0;
+static inline void recordNewCmd(RdbxRespToTcpLoader *ctx, const struct iovec *iov, int iovcnt) {
```
oranagra, line 97: maybe i'm not devoting enough time to read the code to understand it, but either way, i guess it would be better to add a 2 line comment on each function as to what it does, and if needed also document the inputs and outputs.

additionally, in there's some multi-function mechanism / design of command recording, maybe there should be some big doc block at the top of the file that describes it.

oranagra (commented):
skimmed though it and came up with some questions

src/cli/rdb-cli.c
```diff
@@ -0,0 +1,275 @@
+#include <stdio.h>
+#include <stdarg.h>
+#include <stdlib.h>
+#include <string.h>
```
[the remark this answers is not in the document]
moticless: ok. I will take care to rearrange, before merge.

src/ext/readerResp.c
moticless: No. As long as there are no commands, which sent by respToTcpLoader.c, that will cause to get back 
aggregated replies, I think it is better not support it. 

I do need TODO some effort to have better error reporting and reflect the failed command.

src/ext/respToTcpLoader.c
moticless: I wasn't paying attention as i was thinking that i will get back to this code when doing profiling.
(I also had wrong understanding that blocking IO read() might add wait when it receieves less data than requested)

src/ext/respToTcpLoader.c
moticless: Right now it only records. I will need to devote another phase for debugging and testing error flows.

src/ext/readerResp.c
oranagra: at some point or another, we might end up sending a command that does return that.
but anyway, for now, let's add a comment saying why it's dead code..

oranagra (approved):
Skimmed though the changes, LGTM.

PR #8, opened 2023-07-31
Support hash; Fix pointer invalidation due to realloc() by Bulk-pool

moticless:


moticless:
based on wrong branch. Creating a new PR instead.

PR #9, opened 2023-07-31
Add support to HASH and SET datatypes

moticless:
Following this PR, the parser will be able to parse HASH and SET data types. 
* This includes: plain hash, ziplist hash, listpack hash, zipmap hash, plain set, intset and listpack set.
* Added corresponding tests for json, resp and redis formaters.
* Data integrity doesn't check for duplication of elements that was read from RDB (if it is important to client, then it is the one to implement it).
* Added corresponding tests for `json`, `resp` and `redis` formaters.
* As list data-type also uses ziplist and listpack - some refactor was made at `parserRaw.c`.
* Copied implementation of zipmap from Redis to `deps/redis/zipmap.c`

**How to review:** 
* `librdb-api.h`
* tests: review the filenames. Review the json files to see the expected output.
* Please review core parser code (`parser.c` and `parserRaw.c`)

**Another commit:**
fixed `BulkInfo` pointers invalidation due to realloc() by `BulkAlloc.c`. The fix is to avoid any realloc() and use simple list instead. To review it, see `bulkAlloc.c`


src/lib/parserRaw.c
```diff
@@ -9,6 +9,8 @@
 #include "../../deps/redis/util.h"
 #include "../../deps/redis/listpack.h"
 #include "../../deps/redis/ziplist.h"
+#include "../../deps/redis/zipmap.h"
+#include "../../deps/redis/intset.h"
```
oranagra, line 13: unrelated to these lines.. maybe it's a good idea to add a comment at the top of each of the main files explaining their role. i.e. parser.c vs parserRaw.c

src/lib/parser.c
```diff
@@ -840,35 +777,239 @@ static RdbStatus listListpackItem(RdbParser *p, BulkInfo *lpInfo) {
...
+
+        ziplistGet(offsetZL, &item, &itemLen, &itemVal);
+        offsetZL = ziplistNext(ziplistBulk->ref, offsetZL);
+
+        if (!(binfo = allocBulkForEmbeddedString(p, item, itemLen, itemVal, &endChar, &pEndChar)))
+            return RDB_STATUS_ERROR;
+
+        registerAppBulkForNextCb(p, binfo);
+        CALL_HANDLERS_CB(p,
+                         *pEndChar = endChar, // finalize by restore modified char
```
oranagra, line 809: maybe wrapping this assignment in a function would make it more readable?
i.e. calling a function and passing these two vars as arguments.
it'll mean it has a name, and a better place to write a bigger comment without having to repeat it.
i don't recall if i already suggested it, and maybe it's impossible...

src/lib/parser.c
```diff
@@ -840,35 +777,239 @@ static RdbStatus listListpackItem(RdbParser *p, BulkInfo *lpInfo) {
...
+                         items);
     }
+    return RDB_STATUS_OK;
+}
 
-    return nextParsingElement(p, PE_END_KEY);
+RdbStatus hashListPackPairs(RdbParser *p, BulkInfo *lpBulk) {
+    size_t items = 0;
+
+    if (unlikely(0 == lpValidateIntegrity(lpBulk->ref, lpBulk->len, 1, counterCallback, &items))) {
```
oranagra, line 926: so for all ziplists and listpacks we have O(2N)? maybe it should be configurable?
it is in redis...

src/lib/parser.c
```diff
@@ -1206,14 +1354,225 @@ RdbStatus elementQuickList(RdbParser *p) {
...
+
+    /*** ENTER SAFE STATE ***/
+
+    if (RDB_STATUS_ERROR == listZiplistItem(p, ziplistBulk))
+        return RDB_STATUS_ERROR;
+
+    return nextParsingElement(p, PE_END_KEY);
+}
+
+RdbStatus elementHash(RdbParser *p) {
```
oranagra, line 1370: these functions handle the whole key (of a certain encoding), not just a single element.
i.e. they're alliteratively called in steps to consume just a portion each time, but the first portion (the header) is not an element.
consider renaming..

src/lib/parser.c
```diff
@@ -840,35 +777,239 @@ static RdbStatus listListpackItem(RdbParser *p, BulkInfo *lpInfo) {
...
+                         rdbData.handleHashElement,
+                         binfo1->ref,
+                         binfo2->ref,
+                         items);
     }
+    return RDB_STATUS_OK;
+}
 
-    return nextParsingElement(p, PE_END_KEY);
+RdbStatus hashListPackPairs(RdbParser *p, BulkInfo *lpBulk) {
```
oranagra, line 923: i'm not sure the "pairs" part is needed. it's a hash made of a listpack, of course it's pairs.
i think you took the "pairs" part from redis, where the function wasn't type specific.

api/librdb-api.h
```diff
@@ -193,15 +200,22 @@ typedef struct RdbHandlersRawCallbacks {
...
-    RdbRes (*handleListNode)(RdbParser *p, void *userData, RdbBulk node);
-
-    /*** TODO: RdbHandlersStructCallbacks: handlerHashListPack, handleSetIntset, handleZsetListPack, handleFunction ***/
-    RdbRes (*handlerHashLP)(RdbParser *p, void *userData, RdbBulk hashLp);
-    RdbRes (*handleSetIntset)(RdbParser *p, void *userData, RdbBulk intSet);
-    RdbRes (*handleSetLP)(RdbParser *p, void *userData, RdbBulk setLP);
-    RdbRes (*handleSetZL)(RdbParser *p, void *userData, RdbBulk setZL);
+    RdbRes (*handleListLP)(RdbParser *p, void *userData, RdbBulk listLP);
+    /* hash */
+    RdbRes (*handleHashPlain)(RdbParser *p, void *userData, RdbBulk field, RdbBulk value, uint64_t totalNumElm);
```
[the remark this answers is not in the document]
moticless: regarding your suggestion of renaming to `handleHashElement`, I prefer not to use `Element`. Like mentioned above, what about `handleHashLegacy`? It makes (to me) very clear what kind of "structure" to expect.

We can discard variable `totalNumElm` and provide instead API to query for hint. Something like: 
`RDB_getNumItemsHint(RdbParser *p)`

api/librdb-api.h
```diff
@@ -193,15 +200,22 @@ typedef struct RdbHandlersRawCallbacks {
...
+    RdbRes (*handleHashZL)(RdbParser *p, void *userData, RdbBulk listZL);
+    RdbRes (*handleHashLP)(RdbParser *p, void *userData, RdbBulk hashLp);
+    RdbRes (*handleHashZM)(RdbParser *p, void *userData, RdbBulk hashZM);
```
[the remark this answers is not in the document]
moticless: As mentioned above, maybe renaming RdbBulk variables will be sufficient.

src/lib/parser.c
moticless: Not sure i understand. The term `element` is mainly aligned with RDB data-structures and opcodes. In this case elementHash will be called on opcode `RDB_TYPE_HASH` and will know how to parse the hash from start to the end, including its header. It will be done in steps, and yet.

api/librdb-api.h
```diff
@@ -193,15 +200,22 @@ typedef struct RdbHandlersRawCallbacks {
 typedef struct RdbHandlersStructCallbacks {
     HANDLERS_COMMON_CALLBACKS
     RdbRes (*handleStringValue)(RdbParser *p, void *userData, RdbBulk str);
-    RdbRes (*handleListLP)(RdbParser *p, void *userData, RdbBulk listLP);
+    /* list */
+    RdbRes (*handleListPlain)(RdbParser *p, void *userData, RdbBulk node);
```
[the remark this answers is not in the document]
oranagra: i don't think "legacy" is right. it should describe the format, not it's history.
ZL is also legacy, and LP might be some day too.
does this callback handle only the old RDB_TYPE_LIST? or also the QUICKLIST_NODE_CONTAINER_PLAIN item in RDB_TYPE_LIST_QUICKLIST_2?

api/librdb-api.h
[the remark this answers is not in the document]
oranagra: querying the hint with an API sounds good idea to me.

api/librdb-api.h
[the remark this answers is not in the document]
oranagra: i still think some comment describing each callback is a good idea.

src/lib/parser.c
oranagra: in this case i think RDB_TYPE_HASH is a "key" not an "element".
i think we refer to the sub-key "fields" as "elements"
so what i meant is that since this function handles a key (not a single field), it should be renamed.
maybe `handleHashKey` or `handleHashType`?

api/librdb-api.h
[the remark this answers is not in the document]
moticless: It also handles QUICKLIST_NODE_CONTAINER_PLAIN. Maybe we better distinct between the two?

src/lib/parser.c
moticless: I tried to have common naming convention across the "parsing-elements" (see peInfo[] table, at parser.c and in readme.md). That is, the term "element" is heavily used. Note that some of the "elements" in the RDB doesn't represent keys. Maybe we can abbreviate the names to `peHash`, `peQuickList`, etc.

api/librdb-api.h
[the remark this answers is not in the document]
oranagra: AFAIK, the only two use cases for this struct callbacks are:
1. memory usage estimation 
2. populating a db in redis itself

i suppose that either of them, can make all the necessary decision on what to do with the info, even if it uses the same callback (without a need to make a distinction between them).
on the other hand, the higher level set of callbacks is suppose to give you the data when you care about the data, and this lower level set of callbacks are suppose to give you raw data on how it is encoded in the rdb.

bottom line, i don't have a clear opinion on which is better, but only after we decide if we wanna split them we should choose names.
i suppose that if we split them, then it'll be [handleListZL, handleListLP, handleListElementSingle, handleListElementPlain]
if we keep them together, it should be [handleListZL, handleListLP, handleListElement]
WDYT?

src/lib/parser.c
oranagra: so "element" is a generic term for some unit in the rdb, and it can be a key, a field, or anything else (aux hint)? 
it sounds a bit odd to me, but it could be just me.
in which case, many places where i was thinking of "elements" they should be "fields" (in hash) or "members" (in set).
i'm not sure.. i'll leave this up to you.

api/librdb-api.h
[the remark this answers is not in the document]
moticless: I prefer to have it simple for now:  [handleListZL, handleListLP, handleListElement].
Note that I escaped from using the term `handle` since it is already used by application handlers. I used instead `element`. Please consider my note below about it.

src/lib/parser.c
moticless: Actually I borrowed the "element" term from XML terminology. I didn't want to reuse the term `handler` as it is already heavily used by the application handlers. Though, I see your point. Please let's take it on a distinct PR.

api/librdb-api.h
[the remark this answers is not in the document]
oranagra: so "element" is taken and we can't use handleListElement either.
in that case we can use "item" for list, "field", and "member" for set.
that should be reflected in RdbHandlersDataCallbacks too.

i don't mind not using the term "handle", but we must define what's an "element" and if it can also refer to a complete key, we can't use it to indicate a field.
IMHO, i'd rather avoid using it as you do now and reserve it for sub-key items only, but i'm ok with your way as long as we don't also use it for fields.

api/librdb-api.h
```diff
@@ -330,6 +330,19 @@ _LIBRDB_API void RDB_setLogLevel(RdbParser *p, RdbLogLevel l);
 _LIBRDB_API void RDB_setLogger(RdbParser *p, RdbLoggerCB f);
 _LIBRDB_API void RDB_log(RdbParser *p, RdbLogLevel lvl, const char *format, ...);
 
+/* Following function returns a hint for the total number of items in the current
+ * parsed key context - to assist with memory allocation or other optimizations.
+ * Return valid value inside the following callbacks:
```
oranagra, line 335: please note that the existence of the hint depends on the encoding, and is not always available.

moticless: I specified in the comment explictly in which callbacks it will return a valid value. Please lmk if you think it is not so.

oranagra: my argument is that it's not about which states or callbacks will have that data, i think the user should expect to get `-1` anyway (even in these callbacks).
i'm not even sure we need to list in which callbacks it may or may not be available (what we have today may change tomorrow), it's a hint that may or may not exist..

oranagra (approved):

api/librdb-api.h
```diff
@@ -159,75 +166,144 @@ typedef struct RdbStreamConsumerMeta {
...
+
 typedef struct RdbHandlersDataCallbacks {
     HANDLERS_COMMON_CALLBACKS
+    /* Callback to handle the value of a string key */
     RdbRes (*handleStringValue)(RdbParser *p, void *userData, RdbBulk str);
-    RdbRes (*handleListElement)(RdbParser *p, void *userData, RdbBulk str);
+    /* Callback to handle an item from a list */
+    RdbRes (*handleListItem)(RdbParser *p, void *userData, RdbBulk item);
+    /* Callback to handle a field-value pair within a hash */
+    RdbRes (*handleHashFieldValue)(RdbParser *p, void *userData, RdbBulk field, RdbBulk value);
```
oranagra, line 285: i think we can drop the "value" part.
we don't use it in ZSET (although we do have a score)/

oranagra (approved):

PR #12, opened 2023-08-13
Extend rdbloader to support FD

moticless:
- Fixing rdb parsing of empty database
- To be on the safe side, handling of Incomplete writev()
- Extend RespToTcpLoader to support FD
- Rebrand RespToTcpLoader to be RespToRedisLoader.
- Extend RdbxToResp to support flag delKeyBeforeWrite.
- Extend RdbxRespToRedisLoader to know mapping RX RESP error to TX CMD.


api/librdb-ext-api.h
```diff
@@ -158,15 +155,29 @@ _LIBRDB_API RdbxRespFileWriter *RDBX_createRespFileWriter(RdbParser *p,
...
+        } tcp;                          /* TCP connection */
+    } u;
+} RdbxRedisConn;
+
+_LIBRDB_API RdbxRespToRedisLoader *RDBX_createRespToRedis(RdbParser *p,
                                                             RdbxToResp *rdbToResp,
-                                                            const char* hostname,
-                                                            int port,
-                                                            int pipelineDepth);
+                                                            RdbxRedisConn *redisConn);
```
oranagra, line 180: i don't think taking a struct pointer, let alone one with union, is a good interface.
maybe we should expose several "create" functions?
and if they have many common attributes, we can move these to some "set"ter functions?

oranagra: this API is suppose to be externally available, right?
with stable ABI and multi-language binding...

src/ext/handlersToResp.c
```diff
@@ -188,6 +183,16 @@ static RdbRes toRespNewKey(RdbParser *p, void *userData, RdbBulk key, RdbKeyInfo
...
+
+    if (ctx->conf.delKeyBeforeWrite) {
+        struct iovec iov[4];
+        char keyLenStr[32];
+        IOV_CONST_STR(&iov[0], "*2\r\n$3\r\nDEL\r\n$");
+        iov_stringLen(&iov[1], ctx->keyCtx.keyLen, keyLenStr);
+        IOV_STRING(&iov[2], ctx->keyCtx.key, ctx->keyCtx.keyLen);
+        IOV_CONST_STR(&iov[3], "\r\n");
+        return writevWrap(ctx, iov, 4, 1, 1);
+    }
```
oranagra, line 195: maybe we can know the key type at this stage, and if we know it's gonna be handled by SET or RESTORE then we don't need that DEL?

src/ext/respToRedisLoader.c
```diff
@@ -0,0 +1,243 @@
+#include "common.h"
+#include <string.h>
+#include <unistd.h>
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <arpa/inet.h>
+#include <errno.h>
+#include "readerResp.h"
```
oranagra, line 8: since this PR contains a file rename commit, and a refactor rename of its contents, it may be a good idea not to squash-merge it. (it looks like git isn't able to show it as a rename when looking at the whole diff)
in which case maybe you'll want to rearrange the commits to be cleaner before merging

api/librdb-ext-api.h
moticless: Ok, I will split it to `RDBX_createRespToRedisTcp` and `RDBX_createRespToRedisFd`

src/ext/handlersToResp.c
moticless: SET: We do pass the opcode in `RdbKeyInfo *info`. I can add an exception.
RESTORE: Known internally by the parser but not exposed via API (to librdb-extension).

src/ext/respToRedisLoader.c
moticless: Sure. Thanks.

src/ext/handlersToResp.c
oranagra: so since this is a very common use (of the tool, not the parser), and the DELs will make a huge overhead, maybe we wanna handle it?
or maybe we can just make sure that `delKeyBeforeWrite` is off when RESTORE will be used?

moticless: > or maybe we can just make sure that delKeyBeforeWrite is off when RESTORE will be used?

Note that there is an option to multiple levels regsitration at different levels and also function `RDB_handleByLevel` that helps to resolve conflicts such that eventually the parser has a mapping per opcode of which level should be called.

Maybe instead of overloading with another getter function maybe to extend `RdbKeyInfo *info` to pass another argument:
`RdbHandlersLevel handledByLevel`.

src/ext/handlersToResp.c
```diff
@@ -194,11 +202,11 @@ static RdbRes toRespNewKey(RdbParser *p, void *userData, RdbBulk key, RdbKeyInfo
         return RDB_ERR_FAIL_ALLOC;
 
     /* apply del-key-before-write if configured. Not really required for strings */
-    if ((info->opcode != _RDB_TYPE_STRING) && (ctx->conf.delKeyBeforeWrite)) {
+    if ((info->opcode != _RDB_TYPE_STRING) && (ctx->delKeyBeforeWrite == DEL_KEY_BEFORE_BY_DEL_CMD)) {
```
oranagra, line 205: maybe improve the comment to mention why (i.e. that SET is an overwrite command)

oranagra (approved):

oranagra (approved):

PR #13, opened 2023-08-17
rdbloader to support FD. RESP to support delKeyBeforeWrite

moticless:
- Fixing rdb parsing of empty database
- To be on the safe side, handling of Incomplete writev()
- Extend RespToTcpLoader to support FD
- Rebrand RespToTcpLoader to be RespToRedisLoader.
- Extend RdbxToResp to support flag delKeyBeforeWrite.
- Extend RdbxRespToRedisLoader to know mapping RX RESP error to TX CMD.


oranagra (approved):
is that a re-make of the other PR? i.e. with commits organized so that we won't squash-merge it?
you could have just force-push that into the other PR to be merged there, and if you didn't then, let's add a reference so that the discussion can be found.

i'm assuming all the comments in the other PR were handled, so i'm approving it.

moticless:
This PR was aimed to reaarange commits of another PR. 
But eventually force-push was used on original PR.

PR #14, opened 2023-08-28
c99 ; remove dstRdbVersion ; support expire, del before write, Functions

moticless:
Commits:
- Use standard c99 instead of gnu99
- Remove dstRdbVersion from RdbxToRespConf API 
- RDB2RESP: Support EXPIRE and DEL-KEY-BEFORE-WRITE
- Support LRU & LFU
- Extend parser to support Functions.
- Rename respFileWriter to respToFileWriter (to align with respToRedisLoader)
- RDB2RESP: Remove SELECT command exclusion flag from API


api/librdb-ext-api.h
```diff
@@ -93,29 +93,20 @@ _LIBRDB_API RdbxFilterKey *RDBX_createHandlersFilterKey(RdbParser *p,
...
-
-        /* an alternative configuration to dstRdbVersion */
-        char *dstRedisVersion;
-    } restore;
+    /* dstRedisVersion is relevant only if supportRestore is set. It is required
+     * to verify that the version of Redis that will eventually will consume RESP
+     * output of this parser is aligned with the version of the source RDB file.
+     * Otherwise, RESTORE won't be honored and higher level commands will be
+     * generated by the parser instead */
+     const char *dstRedisVersion; /* "<major>.<minor>[.<patch>]" */
```
oranagra, line 109: so since we introduced this thing, we started using it for more than just the decision to send RESTORE or not.
i.e. for ABSTTL, and maybe others in the future.
so let's generalize this comment, and maybe other references to is (like the CLI help message)

src/ext/respToRedisLoader.c
```diff
@@ -123,7 +128,10 @@ static int redisLoaderWritev(void *context, struct iovec *iov, int count, int st
                                     "Failed to write socket. Exceeded EAGAIN retry limit");
                     return 1;
                 }
-                usleep(1000 * retries); /* Backoff and Retries  */
+
+                /* sleep 1msec */
+                struct timespec req = {0, 1000 * 1000}, rem;
+                nanosleep(&req, &rem);
```
oranagra, line 134: sleeps are always evil!!!
why do we have this sleep?
for a plain EAGAIN, we don't need to sleep.
for EWOULDBLOCK maybe we don't want to stay in the loop at all?

src/lib/parser.c
```diff
@@ -1620,7 +1641,7 @@ RdbStatus elementEndOfFile(RdbParser *p) {
         }
     }
 
-    CALL_COMMON_HANDLERS_CB(p, handleEndRdb);
+    CALL_COMMON_HANDLERS_CB_NO_ARGS(p, handleEndRdb);
```
oranagra, line 1644: so the transition to c99 was easy? i'm curious what else did it involve?
nanosleep and static are related?

api/librdb-ext-api.h
```diff
@@ -93,29 +93,20 @@ _LIBRDB_API RdbxFilterKey *RDBX_createHandlersFilterKey(RdbParser *p,
 typedef struct RdbxToRespConf {
     /* TODO: support the option of expire, del, select db */
 
+    int delKeyBeforeWrite;
+    int applySelectDbCmds;  /* if not configured SELECT commands, all keys will be flatten into a single db */
```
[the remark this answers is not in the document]
moticless: Since it is already written & tested, maybe let's just leave it as is. WDYT?

src/ext/handlersToResp.c
```diff
@@ -1,26 +1,45 @@
...
 #define _RDB_TYPE_STRING 0
+#define VER_VAL(major,minor) (((unsigned int)(major)<<8) | (unsigned int)(minor))
+
+typedef struct RedisToRdbVersion {
+    unsigned int redis;
+    unsigned char rdb;
+} RedisToRdbVersion;
+
+const RedisToRdbVersion redisToRdbVersion[] = {
+        {VER_VAL(7,2), 11},
```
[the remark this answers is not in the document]
moticless: sure.

src/ext/respToRedisLoader.c
moticless: Ok I will remove the sleep.

EAGAIN & EWOULDBLOCK are actually the same on modern linux machines:
https://man7.org/linux/man-pages/man3/errno.3.html

src/lib/parser.c
moticless: In addition I had few functions that are not available on c99,
which are nanosleep(), kill(), strdup(), setenv().

At two files I added at the begining the following 
```
# test_common.c
/* feature-test-macros POSIX.1-2008 for: kill(), strdup(), setenv() */
#define _POSIX_C_SOURCE 200809L

# respToRedisLoader.c
/* feature-test-macros POSIX.1-2008 for: nanosleep() */
#define _POSIX_C_SOURCE 200809L
```
If it will bother us in the future we can try eliminate them as well.

moticless: In addition I had few functions that are not available on c99,
which are nanosleep(), kill(), strdup(), setenv().

At two files I added at the begining the following 
```
# test_common.c
/* feature-test-macros POSIX.1-2008 for: kill(), strdup(), setenv() */
#define _POSIX_C_SOURCE 200809L

# respToRedisLoader.c
/* feature-test-macros POSIX.1-2008 for: nanosleep() */
#define _POSIX_C_SOURCE 200809L
```
If it will bother us in the future we can try eliminate them as well.

src/ext/respToRedisLoader.c
oranagra: ohh, i was confusing EAGAIN with EINTER.
then i think we want to loop here only on EINTER, and we want to exit if we get EGAIN or EWOULDBLOCK.

src/lib/parser.c
oranagra: that's perfectly fine.

oranagra (dismissed):

src/cli/rdb-cli.c
```diff
@@ -60,7 +60,7 @@ static void printUsage() {
     printf("\t-l, --log-file <PATH>         Path to the log file (Default: './rdb-cli.log')\n\n");
 
     printf("FORMAT_OPTIONS ('json'):\n");
-    printf("\t-w, --with-aux-values         Include auxiliary values\n");
+    printf("\t-i, --include <EXTRAS>        To include: {aux-val|func}\n");
```
oranagra, line 63: not sure these should be combined in the same option.
is this just for JSON?
i.e. for RESP, the functions are included by default?
in the json we format them in a readable manner, right?

api/librdb-ext-api.h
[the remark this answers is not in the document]
YaacovHazan: I guess we can leave it, but I agree with Oran it doesn't make much sense for a generic tool.

src/cli/rdb-cli.c
moticless: It is just for json part. RESP won't be affected. 
Generally speaking the usage or `rdb-cli` is constructed from two parts:
1. Configuring core parser and common features.
2. Configuring the FORMATTER that will be used.
```
Usage: rdb-cli /path/to/dump.rdb [OPTIONS] {json|resp|redis} [FORMAT_OPTIONS]
```

api/librdb-ext-api.h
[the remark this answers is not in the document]
moticless: Ok. I will drop this change.

src/cli/rdb-cli.c
moticless: Escaping is being applied over the entire json output.

api/librdb-ext-api.h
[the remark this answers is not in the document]
moticless: Reviewing again,I see now that this change requires to add a lot of redundant 
code to the tests which will also will make it less clear. That is, a very simple test 
such as testing SET command outputs, if until now it verified that the output is :
```
*3
$3
SET
$3
xxx
$3
111
```
Now it will verify:
```
*2
$6
SELECT
$1
0
*3
$3
SET
$3
xxx
$3
111
```
Please give it a second thought. Since there is no really additional complexity, easy to grasp, make tests simpler, and gives additional capability out of the box, I prefer to leave it as is.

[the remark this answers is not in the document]
oranagra: you can assume that the default database on connection is db 0, so you'll send SELECT only when changing to another db.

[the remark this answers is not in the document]
moticless: IMHO, this is kind of hidden logic that can be much more risky to users who try to use or extend the parser, and say play around between db0 and db1, but they will end up with unclear scenario that sometimes it calls to SELECT, and sometimes not. Or worse ... wrongly uploading db_num_0.rdb on db 1.

src/lib/parser.c
moticless: @tezc , to workaround it I refactored the *HANDLERS_CB macros to support either with or without arguments.
If you need the very same macros, then do your own temporary change to workaround it and consume this PR once pushed.

IMO, our best effort is to stick to the standard (c99) and if we cannot, then see what are the options to allow it gracefully.

tezc: @moticless I'm not in hurry, I can definitely wait your PR :) IIUC you will push that change to this PR? 

Btw, there are a few more places like `ouput_fprintf(...)`, `UNUSED(...)`. You can change CC = clang and check it if you want.

api/librdb-ext-api.h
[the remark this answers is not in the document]
oranagra: I don't think so, but what do you suggest?
A flag to omit SELECT that's off by default, but used in all tests?
IMHO always emitting SELECT, except for SELECT 0 when we're the ones that establish the connection and can assume db 0 is already selected is ok.

It gets complicated since we're not always the ones that establish the connection.
So maybe as an api, we can expose an optional argument for the caller to tell us which db is selected (or -1). And we keel track of what we changed. 
I.e if the input says db 4 is already selected, and the only db in the rdb is 4, we'll omit select.
And if the input is -1 we won't (even for db 0).

If the case we establish the connection, and also in tests, we can pass 0.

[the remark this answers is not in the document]
moticless: The only reason i felt comfortable to have flag `applySelectDbCmds` in the API, is because behind 
the internal inscentive of using it for the tests, it suggests additional capability to the user. It opens 
the option to play the same RDB file on different db-nums. Or totally ignore the existance of 
db-nums in Redis.

Your last suggestion of passing selected db-num , offers nothing to the user. It just discard single
SELECT command if we are already on the same db-num. Which makes less sense to have it in 
the API.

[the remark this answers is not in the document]
oranagra: i suggested it for two reasons:
1. make the majority of the tests shorter
2. if we're establishing the connection, we can assume db0 is already selected.

i personally don't think we need to offer any such option to the user. if we keep it internal just for testing, then it's probably fine.
i personally don't think the tool should support flattering multi-db into one as a feature.
anyway, the discussion become too long, so do what you feel is right.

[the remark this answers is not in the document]
moticless: ok. I removed the flag :). 

Instead, for testing purposes alone, I've defined custom filter handlers that do not pass new-db callbacks to the RDB to RESP handlers. Thanks.

oranagra (approved):

PR #15, opened 2023-09-04
Fix multiple function definitions

tezc:
Build fails when compiled with optimization level `-O0`.  
`-O3` strips symbols and there is no clashes but `-O0` keeps all definitions and makes the build fail.

### DIFF
--- a/test/test_common.c
+++ b/test/test_common.c
@@ -167,7 +167,6 @@ void cleanupRedisServer() {
 }
 
 void setupRedisServer() {
-    int status = 0;
     pid_t pid = fork();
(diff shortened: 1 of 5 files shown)


test/test_common.c
```diff
@@ -167,7 +167,7 @@ void cleanupRedisServer() {
 }
 
 void setupRedisServer() {
-    int status;
+    int status = 0;
```
moticless, line 170: looks like `status` is not used at all. Can be dropped.

moticless (dismissed):

moticless (approved):

PR #16, opened 2023-09-04
Fix parser state debug logging

tezc:


moticless (approved):
Better to aggregate small changes into one PR. Easier to read history.

moticless:
@tezc, 
IMHO, when we group related changes together into one commit, it creates a much cleaner and more organized commit history to read. Better to aggregate multiple minor changes into a single commit. Obviuosly important changes should have a distinct commit.

PR #17, opened 2023-09-05
Support parsing modules

moticless:
- Support parsing modules.
- Fix bug in rdbLoadLen() when it reads 64bits
- Support parsing module-auxiliary data 
- Adding deps/redis/config.h for BYTE_ORDER correctness
- Add hiredis as submodule for testing
- Utilize hiredis for the tests
- RDB2RESP: Extend API with flag supportRestoreModuleAux


src/lib/parser.c
```diff
@@ -1678,8 +1682,103 @@ RdbStatus elementFunction(RdbParser *p) {
...
+            }
+            default:
+                RDB_reportError(p, RDB_ERR_MODULE_INVALID_STATE,
+                    "elementModule() : Invalid parsing element state: %d.", ctx->state);
+                return RDB_STATUS_ERROR;
+        }
+    }
+
+    /* Static assert that verifies enums are aligned */
+    _Static_assert(ST_MODULE_OPCODE_SINT==RDB_MODULE_OPCODE_SINT, "enums must be aligned");
```
tezc, line 1759: `_Static_assert` was introduced in c11 if I'm not mistaken. gcc is generally doing terrible about warning these issues but clang would complain most probably.

moticless: Let's refactor such kind of code when we goanna have CI that support clang to verify it.

tezc: Just fyi, In my experience, compilers don't warn you reliably about the standard (it should be an easy thing to do though). Generally, you don't know until compile fails on some platform. So, I try to avoid features if I know it does not exist in a standard or on a platform. 

Anyway, here it is clang build, maybe it will throw some warnings: https://github.com/redis/librdb/pull/18

src/lib/parserRaw.c
```diff
@@ -583,6 +583,119 @@ RdbStatus elementRawSet(RdbParser *p) {
...
+        ST_RAW_MODULE_GET_ID=0,
+
+        /* Following enums are aligned to module-opcodes. Static assert below. */
+        ST_RAW_MODULE_OPCODE_SINT,
+        ST_RAW_MODULE_OPCODE_UINT,
+        ST_RAW_MODULE_OPCODE_FLOAT,
+        ST_RAW_MODULE_OPCODE_DOUBLE,
+        ST_RAW_MODULE_OPCODE_STRING_CALL_STR,
+
+        ST_RAW_MODULE_OPCODE_STRING_STR_RETURN, /* return from sub-element string parsing */
```
oranagra, line 597: i don't understand (yet) this mechanism of CALL_STR and STR_RETURN.
maybe it's explained in code i didn't yet get to.

src/lib/parserRaw.c
```diff
@@ -583,6 +583,119 @@ RdbStatus elementRawSet(RdbParser *p) {
...
+                return subElementCall(p, PE_RAW_STRING, ST_RAW_MODULE_OPCODE_STRING_STR_RETURN);
+            }
+
+            case ST_RAW_MODULE_OPCODE_STRING_STR_RETURN: {
+                /*** ENTER SAFE STATE (no rdb read)***/
+                size_t len;
+                char *dataRet;
+                subElementCallEnd(p, &dataRet, &len);
+                updateElementState(p, ST_RAW_MODULE_NEXT_OPCODE);
+                break;
```
oranagra, line 662: what's that safe state? safe for what and when does it end?

test/dumps/module_data.json
```diff
@@ -0,0 +1 @@
+"key1":"<<BINARY>>"
```
oranagra, line 1: maybe we can print the module name and maybe size of the payload?

test/test_common.c
```diff
@@ -216,19 +206,38 @@ void setupRedisServer() {
...
+        /* if module testrdb.so exists (ci.yaml takes care to build testrdb), part
+         * of redis repo testing, then load it for test_rdb_to_redis_module. The
+         * test will run only if testrdb appear in the server "MODULE LIST",
+         * otherwise skipped gracefully. */
+        if (access(testrdbModulePath, F_OK) != -1) {
+            execl(fullpath, fullpath,
+                  "--port", redisPortStr,
+                  "--dir", "./test/tmp/",
+                  "--logfile", "./redis.log",
+                  "--loadmodule", testrdbModulePath,
```
oranagra, line 226: do we have a test that uses RESTORE command and validates the module data was loaded on the dest correctly?

src/lib/parserRaw.c
moticless: Available for a while and exaplained in the code next to function `subElementCall` and struct `ParsingSubElement` :
```
/****************************************************************
 * Sub-Element parsing
 *
 * The parser can handle one level of nested parsing-elements (PE), whereby a PE
 * may be called by another PE and control is returned to the caller once the
 * parsing of sub-element is complete.
 *
 * Currently, this functionality is only adapted and utilized by raw string element
 * which can also run as sub-element of other data-types. It takes care to write the
 * processed string to raw-aggregator, as part of RESTORE of current command.
 *
 * In addition it returns the string decoded by allocating bulkUnmanagedAlloc()
 * (The caller won't need to release returned data yet restrictd to use it within
 * its current parsing-element state).
 ****************************************************************/
```

src/lib/parserRaw.c
moticless: Available for a while and exaplained in the README. Please see section `Parsing-Element states` and subsection ` Defining Safe-state`. Please let me know if it is clear enough.

test/dumps/module_data.json
moticless: Maybe we can optionally return some kind of iterator (for int, float, double, double & string) and let handlers to decide what they want to do with it ... But then the handlers might need to deal with `RDB_STATUS_WAIT_MORE_DATA`. If so, then it will be simpler to have another group of callbacks to submodules (that called only after each item was entirely read) and let the handlers do the math, if it wants the size.

test/test_common.c
moticless: See function `test_rdb_to_redis_module`. It might be little fragile though. As precondition, the test verifies that testrdb module is loaded into redis, otherwise the test gracefully skipped.

(Note that all quick and dirty `redis-cli` system calls should be replaced with integrtaing `hiredis` library eventually)

src/lib/parserRaw.c
oranagra: so for module data types, you this sub-element for strings, and but not for other primitives?
is that described somewhere?
i understand it let's you break big module payloads to many individual calls (even split the string to several calls and avoid blocking IO), right?
what would happen if a module payload is composed of a ton of integers (let's say 100mb worth of integers)?

test/dumps/module_data.json
oranagra: IIUC, you're suggesting that users can implement extensions to the parser that can decode module keys and translate them to usable content (e.g. for ReJson, they'll be able to print the JSON content of the key).
that sounds like a good idea, maybe we can add that in a todo and handle it in a second phase.

meanwhile, for someone that dumps an RDB into json and some keys are module keys, i think just saying "BINARY" is insufficient, and i'd like to state the module name and maybe the payload size.

test/test_common.c
oranagra: great.. why skip it if the module load fails? i'm afraid we'll miss it instead of fixing the problem / improving the test.

src/lib/parser.c
moticless: As there is a lot of vudu code to have cross-version code for assert, maybe we can limit static-assert just for some of the versoins which is good enough.

src/lib/parserRaw.c
moticless: Right.

It will call inefficiently to string-element multiple times through `parserMainLoop()`. Those inefficient calls can be optimized by making direct call and avoid from passing through `parserMainLoop()`. For example, instead of : 
```
            case ST_RAW_MODULE_OPCODE_STRING_CALL_STR: {
                /* call raw string as subelement */
                return subElementCall(p, PE_RAW_STRING, ST_RAW_MODULE_OPCODE_STRING_STR_RETURN);
```

We shall have:
```
            case ST_RAW_MODULE_OPCODE_STRING_CALL_STR: {
                /* call raw string as subelement */
                subElementCall(p, PE_RAW_STRING, ST_RAW_MODULE_OPCODE_STRING_STR_RETURN);
                IF_NOT_OK_RETURN(elementRawString(p)); 
                break;   /* short transition between module states without passing through main-loop */

```

I should reach later to performance phase. Yet to plan what is goanna be tested.

moticless: Maybe I missed your point. If it is native module integer then it won't call to string and the transition between the states of the module will be made only within element-module boundary (i.e. function `elementRawModule()`)

test/test_common.c
moticless: ci.yml should fail if failed to build `testrdb.so`. If we reach here and the file exists then for sure the test will capture the fail.
The graceful fallback to the case `testrdb.so` doesn't exists is to be able, later on, to run these tests against earliest redis version that I can reach. In addition there is the option to expose the tests for external redis server which can assist debugging Redis side in the future.

src/lib/parserRaw.c
oranagra: my last question was if it'll break out in between module integers and return to the caller?
and for string we use the sub-element in order to break if further (i.e. break the string into smaller chunks)?

test/test_common.c
oranagra: it's a little odd to depend on the module from redis, and even odder that it's loosely based on it (graceful failures).
maybe we should bring that module (with some simplifications) into this repo and avoid that dependency.

src/lib/parser.c
tezc: not sure I understand. Do you want to keep static assert? If so, this is how it is defined on my platform in cdefs.h: 

```c
#define _Static_assert(expr, diagnostic) \
    extern int (*__Static_assert_function (void)) \
      [!!sizeof (struct { int __error_if_negative: (expr) ? 2 : -1; })]
```

I guess you can define this if undefined. 


Btw, maybe you can assign those defines directly and avoid checking it altogether:

```c
enum {
    ST_MODULE_OPCODE_SINT = RDB_MODULE_OPCODE_SINT 
    ...
}
```

oranagra: i wanna remind us that these values are part of the rdb format and are never gonna change.
it's not like some other (volatile-only) enum, that can change but two components must remain allied.
it's like asserting that TCP is protocol number 6 in the IP packet header.
not needed IMHO.

moticless: @tezc, 
I wrote it first like that but then I thought about the issue that enum doesn't verifies duplication of values.
But then thinking about it again, it might be good compromise between readability of the code and alignment enforcement. I think it I will go with your suggestion. 

@oranagra,
The verification is around changing the states values of the element that might break the alignment with the rdb opcodes. Not to verify that rdb opcodes has suddenly changed.

src/lib/parserRaw.c
moticless: Each new (sub)element+state is being updated and can reinvoked in case of early return. 
Most of the librdb tests run also in a mode that fails all "new" read and simulates RDB_STATUS_WAIT_MORE_DATA from RDB reader/buffer, which enforces the parser to return from its current "state" and reinvoke again.

src/ext/handlersToResp.c
```diff
@@ -380,73 +382,131 @@ static RdbRes toRespFunction(RdbParser *p, void *userData, RdbBulk func) {
...
+    /* reset rawCtx */
+    RdbxToResp *ctx = userData;
+    ctx->rawCtx.rawSize = size;
+    ctx->rawCtx.sentFirstFrag = 0;
+    ctx->rawCtx.isModuleAux = 1;
+    ctx->rawCtx.crc = 0;
+
+    /* if target doesn't support module-aux, then skip it */
+    if (!ctx->conf.supportRestoreModuleAux)
+        return RDB_OK;
```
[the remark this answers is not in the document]
moticless: It will add more complexity to responses pipeline tracking that I am not sure if it is worth it.
Another option is to extend boolean variable `supportRestore` to be:
 0=No support, 1=Support restore, 2=Support Restore with Module Aux

In the tradeoff of spamming the API with weird flags and deep hiding such feature, this sound 
most reasonable to me.

src/ext/handlersToResp.c
```diff
@@ -380,73 +382,131 @@ static RdbRes toRespFunction(RdbParser *p, void *userData, RdbBulk func) {
 
 /*** Handling raw ***/
 
+static RdbRes toRespRawBeginModuleAux(RdbParser *p, void *userData, RdbBulk name, int encver, int when, size_t size) {
```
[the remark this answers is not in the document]
moticless: Few lines below it can be seen assigned to `ctx->rawCtx.rawSize = size;`
I will modify it to `rawSize`

src/ext/handlersToResp.c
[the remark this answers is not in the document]
oranagra: i prefer not to add an interface at all (just silently ignore), but if we don't want that, then a separate interface would be easier to remove later (rather than mixing it with another interface).

i think at some point you'll have to have the complexity of understanding responses and associating them with the commands you sent anyway, and once you do that, then this optimistic approach won't add a big overhead.

one example is that when a command fails, i think we must log which command it was, not just the error message, otherwise, trouble shooting will be very difficult.

src/ext/handlersToResp.c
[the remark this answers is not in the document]
oranagra: but how do we know the rawSize when we begin a module aux? where do you take that info from?
i have a feeling (didn't bother to look), that it's always 0, and not used in this case...

src/ext/handlersToResp.c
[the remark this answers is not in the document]
moticless: Note that librdb can generate RESP output. How we suppose to decide whether to generate this special command or not? in that case, I prefer to leave the flag as is.

BTW, I do have some basic mechanism for pipeline to report of which command caused the error. Example to 100 pending list commands and failure on `mylist62` because the test makes "SET mylist62 1" before
```
:: ERROR :: [elementQuickList::State=1] 
Received Server error: "WRONGTYPE Operation against a key holding the wrong kind of value"
Got failed on command [#9501] (First 40 bytes):
*3
$5
RPUSH
$8
mylist62
$19
0.035

Preceding command [#9500] was: 
*3
$5
RPUSH
$8
mylist30
$17
0.791
```

src/ext/handlersToResp.c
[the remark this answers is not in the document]
moticless: Please read the comment:
```
typedef struct RawContext {

    /* aggType - The parser encounters difficulty determining the size of
     * certain types during their parsing because they lack a preceding header
     * that would indicate their size. This creates a problem when implementing
     * callback handlers for executing RESP2 RESTORE commands on a live server,
     * which requires the type's size to be sent at the beginning. To resolve
     * this issue, the parser gathers all payload data for these types and only
     * provides it to callback handlers once it reaches the end of the type and
     * knows its total size. However, for types like strings, whose sizes are
     * already known at the beginning, the parser will not aggregate the entire
     * payload if it is large enough, but stream it to handleFrag callback. */
    enum {
        AGG_TYPE_UNINIT,
        AGG_TYPE_ENTIRE_DATA,
        AGG_TYPE_PARTIALLY
    } aggType;
```

src/ext/handlersToResp.c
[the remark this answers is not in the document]
oranagra: good point about sending RESP to a file rather than redis.
but then i wonder what do people do with that file? if they pipeline it to redis later (e.g. with `netcat`), then maybe they ignore replies (and errors) anyway.

regarding sending it to a live server, good that you have that feature in the error report.. but that means that it's not too complicated to have it implicitly ignore some errors (like unknown command, only on a certain command).

anyway, my point is that at some point in time we'll want to drop that interface, so i rather not have it if we can avoid it.

src/ext/handlersToResp.c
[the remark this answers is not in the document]
oranagra: ok, i didn't remember the difference between this parser and the one i was involved with.
i.e. in the other parser the aggregation was on the receiver's side, so when starting a new payload the size wasn't known.

So here, it aggregates the entire thing before calling "Begin", and then it repeatedly calls the "Frag" callback? (i suppose that's necessary since each of them may be sitting in a separate allocation).

so this one doesn't have the optional size hint callback, and the "First" frag concept is internal to the callee, right?

thanks for clearing this up...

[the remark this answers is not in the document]
moticless: Exactly

test/test_common.c
moticless: Note that compiling module requires to have redis source. I prefer to leave it as is quick and dirty (?!). And when we reach to a point that we want to expand the support in modules, then make the effort. 
Note that graceful failure is when the `testrdb.so` is not available. In our CI I will take care that it will verify success of `make -C tests/modules`

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "deps/hiredis"]
+	path = deps/hiredis
+	url = https://github.com/redis/hiredis.git
```
tezc, line 3: Maybe not part of this PR but do you think we can use subtree instead of submodule? With subtree, dependency project code is always in the repo, no need to run extra commands to fetch/clear it. (As I mentioned before, I also feel the same way for cmocka as well, it should be part of the repo to make the developers' life easier)

moticless: I am not much familiar with subtree. Sound like a good idea. Let's do it on a distinct PR.

test/test_common.c
oranagra: compiling a module does NOT require the redis source, only redismodule.h, which is commonly just copied into the module's source tree.
i think your test better be self-sufficient and have their own module (copied from redis and cleaned).
you can leave that for a later step, but i think that's the right approach.

moticless: I agree with you that this is the right approach... I am too lazy about it because I know we goanna  expand it and write it properly (It can be valuable feature for modules)

YaacovHazan (approved):

YaacovHazan (approved):

YaacovHazan (approved):

YaacovHazan (approved):

YaacovHazan (approved):

YaacovHazan (approved):

YaacovHazan (approved):

.github/workflows/ci.yml
```diff
@@ -25,17 +27,18 @@ jobs:
           ctest
           sudo make install
           # Install redis for tests against external Redis server
-          cd ~
-          git clone https://github.com/redis/redis.git
-          cd redis 
-          make -j 4
-
+                              
+          # Build Redis to run it against librdb tests
```
yossigo, line 31: @moticless You may want to consider getting a Redis binary, like we do [here](https://github.com/RedisLabs/memtier_benchmark/blob/4203084b085e0f93bb9130c281362ea651d2140c/.github/workflows/ci.yml#L29). It's potentially faster.

yossigo (approved):

.github/workflows/ci.yml
moticless: Good reference. thanks.

Currently librdb is relying on Redis repo for building redis binary and reusing `testrdb.c` source to test modules parsing.
I know it is preferable to librdb have its own source to build testable module but i prefer to wait with that effort since I plan anyway (second phase?!) to expand callback module API to parse primitives as well (and not only RESTORE module).
All in all, additional  +1minute of CI time.

PR #20, opened 2023-09-20
Add sorted set support

tezc:


api/librdb-api.h
```diff
@@ -262,6 +265,8 @@ typedef struct RdbHandlersStructCallbacks {
 
     /*** TODO: RdbHandlersStructCallbacks: ***/
 
+    /* Callback to handle an item from a plain sorted set */
+    RdbRes (*handleZsetPlain)(RdbParser *p, void *userData, RdbBulk item, double score);
     /* Callback to handle a ziplist-based sorted set value */
     RdbRes (*handleZsetZL)(RdbParser *p, void *userData, RdbBulk ziplist);
     /* Callback to handle a listpack-based sorted set value */
```
moticless, line 272: Please move all the supported zset callbacks above the line : `/*** TODO: RdbHandlersStructCallbacks: ***/`

deps/redis/util.c
```diff
@@ -202,3 +205,99 @@ int lpStringToInt64(const char *s, unsigned long slen, int64_t *value) {
...
+
+
+/* Convert a string into a double. Returns 1 if the string could be parsed
+ * into a (non-overflowing) double, 0 otherwise. The value will be set to
+ * the parsed value when appropriate.
+ *
+ * Note that this function demands that the string strictly represents
+ * a double: no spaces or other characters before or after the string
+ * representing the number are accepted. */
+int string2d(const char *s, size_t slen, double *dp) {
```
moticless, line 217: I wonder why did you choose to use string2d() and not follow redis imp at `rdbLoadDoubleValue()` with `sscanf(buf, "%lg", val)`.

src/lib/parser.c
```diff
@@ -1664,6 +1674,191 @@ RdbStatus elementSetLP(RdbParser *p) {
...
+            RDB_reportError(p, RDB_ERR_PLAIN_ZSET_INVALID_STATE,
+                            "elementZset(): invalid parsing element state: %d", ctx->state);
+            return RDB_STATUS_ERROR;
+    }
+}
+
+/* return either RDB_STATUS_OK or RDB_STATUS_ERROR */
+static RdbStatus zsetZiplistItem(RdbParser *p, BulkInfo *ziplistBulk) {
+
+    int ret = ziplistValidateIntegrity(ziplistBulk->ref, ziplistBulk->len, p->deepIntegCheck, NULL, NULL);
```
moticless, line 1735: please move the location of zsetZiplistItem() upward to be between `/*** various functions ***/` and `/*** Parsing Common Elements ***/`

src/lib/parser.c
```diff
@@ -1664,6 +1674,191 @@ RdbStatus elementSetLP(RdbParser *p) {
...
+
+    /*** ENTER SAFE STATE ***/
+
+    /* Doesn't check for duplication */
+    if (!lpValidateIntegrity(listpackBulk->ref, listpackBulk->len, p->deepIntegCheck, NULL, 0)) {
+        RDB_reportError(p, RDB_ERR_ZSET_LP_INTEG_CHECK, "elementZsetLP(): LISTPACK integrity check failed");
+        return RDB_STATUS_ERROR;
+    }
+
+    /* TODO: handle empty listpack */
```
moticless, line 1817: I tested that redis 7.x doesn't save to rdb zset with 0 members. Please double check
It is the same for old version and if so, please delete this comment. Otherwise, let's
think how to change the code to support it.

src/lib/parser.c
```diff
@@ -1831,8 +2026,9 @@ RdbStatus rdbLoadFloatValue(RdbParser *p, float *val) {
 }
 
 RdbStatus rdbLoadDoubleValue(RdbParser *p, double *val) {
-    BulkInfo *binfoUnused;
-    IF_NOT_OK_RETURN(rdbLoad(p, sizeof(*val), RQ_ALLOC_REF, (char *) val, &binfoUnused));
+    BulkInfo *binfo;
+    IF_NOT_OK_RETURN(rdbLoad(p, sizeof(*val), RQ_ALLOC, NULL, &binfo));
+    *val = *((double*) binfo->ref);
     memrev64ifbe(val);
```
moticless, line 2032: If you intend to push my fix, then please do it also for `rdbLoadFloatValue()`

src/lib/parserRaw.c
```diff
@@ -572,6 +572,97 @@ RdbStatus elementRawSet(RdbParser *p) {
...
+
+RdbStatus elementRawZset(RdbParser *p) {
+    enum RAW_ZSET_STATES {
+        ST_RAW_ZSET_HEADER=0,
+        ST_RAW_ZSET_READ_MEMBER_STR,
+        ST_RAW_ZSET_READ_MEMBER_STR_RETURN,
+        ST_RAW_ZSET_READ_SCORE_DOUBLE,
+        ST_RAW_ZSET_READ_SCORE_STR,
+        ST_RAW_ZSET_READ_SCORE_STR_RETURN,
+    };
```
moticless, line 591: this is probably the best place to explain the state-flow , next to each state. For reference see comments at  `enum RAW_LIST_STATES`.

test/dumps/plain_zset_2_v11_data.json
```diff
@@ -0,0 +1,6 @@
+"redis-ver":"255.255.255",
+"redis-bits":"64",
+"ctime":"1694535571",
+"used-mem":"965512",
+"aof-base":"0",
+"myzset":{"three":"3.33000000000000007","two":"2","one":"1.1100000000000001"}
```
moticless, line 6: I think zset is probably the best place to test our 64bit read/write, values mainuplation , double2str  and str2double.
Please let's come up with more challenging values, such as,  close to (negative) infinity , close to (negative) zero. and (negative) randomly in the middle that will fail on big-little endian issues.

moticless (commented):
nice work!

api/librdb-api.h
tezc: done

src/lib/parser.c
tezc: done

src/lib/parser.c
tezc: Deleted. I already tried it and this is just a copy-paste leftover.

src/lib/parser.c
tezc: done

src/lib/parserRaw.c
tezc: done

deps/redis/util.c
tezc: Initially, added this one for reading from listpack. zset in redis uses `strtod()` while reading double values from it. 

I guess I didn't pay attention to content of `rdbLoadDoubleValue()`. Now, I realize another issue, it is actually doing more than reading strings. Strings have one byte prefix! :fearful: . Not sure how tests worked with the old format. I'll get this one fixed and will use same function with redis. Thanks.

tezc: should be fixed now.

deps/redis/util.h
```diff
@@ -18,5 +18,7 @@ int ll2string(char *s, size_t len, long long value);
 int ull2string(char *s, size_t len, unsigned long long value);
 int lpStringToInt64(const char *s, unsigned long slen, int64_t *value);
 unsigned int getEnvVar(const char* varName, unsigned int defaultVal);
+double zzlStrtod(unsigned char *vstr, unsigned int vlen);
+int d2string(char *buf, size_t len, double value);
```
tezc, line 22: Fyi, these two functions are used for double <-> string conversion in redis.

`zzlStrtod()` is used when you get double out of sorted set. Although it is a small function, I just copied it exactly just as a reminder that this was copied. One issue is it doesn't do error handling internally, I guess this is not a big problem as this is how it works right now in redis.

src/lib/parser.c
```diff
@@ -2036,7 +2037,7 @@ RdbStatus rdbLoadDoubleValue(RdbParser *p, double *val) {
 
 /* Try to read double value and then copy it to the destination including one
  * byte prefix. See rdbLoadDoubleValue() for details. */
-RdbStatus rdbLoadDoubleValueToDest(RdbParser *p, char *dst, int *written) {
+RdbStatus rdbLoadDoubleValueToBuff(RdbParser *p, char *dst, int *written) {
```
moticless, line 2040: Please change `dst` to `buff`

src/lib/parser.h
```diff
@@ -445,7 +445,7 @@ void subElementCallEnd(RdbParser *p, RdbBulk *bulkResult, size_t *len);
 RdbStatus rdbLoadFloatValue(RdbParser *p, float *val);
 RdbStatus rdbLoadBinaryDoubleValue(RdbParser *p, double *val);
 RdbStatus rdbLoadDoubleValue(RdbParser *p, double *val);
-RdbStatus rdbLoadDoubleValueToDest(RdbParser *p, char *dst, int *written);
+RdbStatus rdbLoadDoubleValueToBuff(RdbParser *p, char *dst, int *written);
```
moticless, line 448: Please change `dst` to `buff`

tezc: ok done

src/lib/parser.c
tezc: ok done

moticless (approved):

PR #21, opened 2023-09-20
Do not use empty function declarations

tezc:
New C standard won't allow empty declarations. Looks like `clang-15` already warns about that. 
This PR fixes the warnings.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
   build-clang:
     runs-on: ubuntu-latest
     env:
-      CC: clang-15
+      CC: clang
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -45,7 +45,7 @@ jobs:
   build-clang:
     runs-on: ubuntu-latest
     env:
-      CC: clang
+      CC: clang-15
```
moticless, line 48: We can make the modification to the code and add 'void' but I am not sure it is good idea to bind specific varsion of clang. There are pros and cons of selecting new/old clang (unless we choose both and taking care they are available on the machine and install it if not) from that point IMHO we better leave whatever is on the machine.

tezc: ok, then we should remember this issue (clion already highlights it in IDE).

moticless (approved):

PR #22, opened 2023-10-08
Support: stream, v12, -fvisibility,  CI 5.0 6.0 6.2, make debug. Fix json enc

moticless:
- Support stream
- Extend readerResp to be able reading bulk array (for Stream).
- RDB2JSON: Fix encoding from \x## to standard \u00##
- Make "-fvisibility=hidden" and static link rdb-cli with librdb
- Run CI against redis 5.0,6.0,6.2 and unstable
- Fix Module RESTORE for earlier redis versions
- Added makefile rule "debug" to build with -O0
- Make iovec macros more clear.
- Support latest rdbversion 12.


# Stream support
```
git:(add-stream-support) rdb-cli ../redis/dump.rdb json               
[{
    "my_key":"Hello, Redis!",
    "my_set":["member1","member2"],
    "my_zset":{"Bob":"5","Alice":"10","Charlie":"15"},
    "my_hash":{"field1":"value1","field2":"value2"},
    "my_list":["item1","item2","item3"],
    "mystream":{
      "entries":[ 
        { "id":"1695649068107-0", "items":{"message":"Message1"} },
        { "id":"1695649068110-0", "items":{"message":"Message2"} },
        { "id":"1695649069139-0", "items":{"message":"Message3"} },
        { "id":"1695649446276-0", "items":{"message":"Message4"} },
        { "id":"1695649456516-0", "items":{"message":"Message5"} },
        { "id":"1695893015933-0", "items":{"field1":"value1", "field2":"value2", "field3":"value3"} }
   ]}
}]

git:(add-stream-support) rdb-cli ../redis/dump.rdb json -i stream-meta                                                                                                                   
[{                                                                                                                                                                                                         
    "my_key":"Hello, Redis!",                                                                                                                                                                              
    "my_set":["member1","member2"],                                                                                                                                                                        
    "my_zset":{"Bob":"5","Alice":"10","Charlie":"15"},
    "my_hash":{"field1":"value1","field2":"value2"},
    "my_list":["item1","item2","item3"],
    "mystream":{
      "entries":[ 
        { "id":"1695649068107-0", "items":{"message":"Message1"} },
        { "id":"1695649068110-0", "items":{"message":"Message2"} },
        { "id":"1695649069139-0", "items":{"message":"Message3"} },
        { "id":"1695649446276-0", "items":{"message":"Message4"} },
        { "id":"1695649456516-0", "items":{"message":"Message5"} },
        { "id":"1695893015933-0", "items":{"field1":"value1", "field2":"value2", "field3":"value3"} }],
      "length": 6, 
      "entriesAdded": 6, 
      "firstID": "1695649068107-0", 
      "lastID": "1695893015933-0", 
      "maxDelEntryID": "0-0",
      "groups": [
        {"name": "groupA", "lastid": "1695649446276-0", "entriesRead": 4,
         "pending": [ 
           { "sent": 1695649446276, "id":"1695649446276-0", "count": 1 }],
         "consumers": [
           { "name": "consumerA2", "activeTime": 1696708503339, "seenTime": 1696708503339,
             "pending": [
               {"id":"1695649446276-0"}]}]},
        {"name": "groupB", "lastid": "1695649069139-0", "entriesRead": 3,
         "pending": [ 
           { "sent": 1695649069139, "id":"1695649069139-0", "count": 1 }],
         "consumers": [
           { "name": "consumerB1", "activeTime": 1696708503339, "seenTime": 1696708503339,
             "pending": [
               {"id":"1695649069139-0"}]}]}]}
}]

git:(add-stream-support) rdb-cli ../redis/dump.rdb resp 
*2   
$6    
SELECT  
$1     
0              
*3     
$3     
SET   
$6      
my_key
$13   
Hello, Redis!
*3      
$4      
SADD  
$6             
my_set         
$7     
member1        
*3      
$4
SADD    
$6  
my_set
$7      
member2   
....
```
# Make "-fvisibility=hidden
Set the default visibility of symbols to be hidden. 
This raises an issue with `deps/redis` files because it is currently being 
build internally and staticaly linked with librdb.so whereas librdb-ext.so need it 
as well for few APIs, such as the use of the rax data structure or CRC.

Instead of refactoring and creating an additional shared library, or making
copies for each library (with all the associated implications) we shall expose
specific symbols from librdb.so that are needed by librdb-ext.so. These symbols
won't be documented in the API, but they will be available for use. Since we
compile by default with hidden visibility, we need to mark them explicitly as
visible using the following macro inside `deps/redis`:
```
#define _LIBRDB_HIDDEN_API __attribute__((visibility("default")))
```

While it might initially seem cumbersome to manually designate each function as
visible, this approach has its benefits. By carefully selecting the functions
to expose, we are encouraged to contemplate what we are exposing and why.
Moreover, it helps prevent the cluttering clients of librdb with unnecessary
symbols.

## List of all symbols (All Undefined symbols were filtered-out. Functions of standard libraries):
```
➜  push-to-librdb git:(add-stream-support) nm -D ./lib/librdb.so | grep ' [TDW] ' 
000000000000a1e0 T RDB_alloc
000000000000a310 T RDB_bulkClone
00000000000033b0 T RDB_bulkCopyClone
0000000000003300 T RDB_bulkCopyFree
000000000000a210 T RDB_bulkLen
000000000000a9b0 T RDB_createHandlersData
000000000000a7c0 T RDB_createHandlersRaw
000000000000a8c0 T RDB_createHandlersStruct
00000000000091a0 T RDB_createParserRdb
000000000000a130 T RDB_createReaderRdb
0000000000009610 T RDB_deleteParser
000000000000a790 T RDB_free
000000000000a630 T RDB_getBytesProcessed
000000000000a6f0 T RDB_getErrorCode
000000000000a720 T RDB_getErrorMessage
000000000000ab70 T RDB_getLibVersion
000000000000abc0 T RDB_getMaxSuppportRdbVersion
000000000000a6c0 T RDB_getNumHandlers
000000000000a590 T RDB_getNumItemsHint
000000000000a660 T RDB_getRdbVersion
000000000000a690 T RDB_getState
000000000000aaa0 T RDB_handleByLevel
000000000000a430 T RDB_IgnoreChecksum
000000000000a290 T RDB_isRefBulk
000000000000a490 T RDB_log
00000000000096e0 T RDB_parse
0000000000009ee0 T RDB_parseBuff
000000000000a3c0 T RDB_pauseParser
000000000000a750 T RDB_realloc
0000000000009fd0 T RDB_reportError
000000000000a5f0 T RDB_setDeepIntegCheck
000000000000a400 T RDB_setLogger
000000000000a5c0 T RDB_setLogLevel
000000000000a460 T RDB_setMaxRawSize
000000000000a390 T RDB_setPauseInterval
➜  push-to-librdb git:(add-stream-support) nm -D ./lib/librdb-ext.so | grep ' [TDW] '
0000000000005970 T RDBX_attachRespWriter
00000000000069b0 T RDBX_createHandlersFilterDbNum
0000000000006670 T RDBX_createHandlersFilterKey
0000000000006960 T RDBX_createHandlersFilterType
00000000000071a0 T RDBX_createHandlersToJson
0000000000003880 T RDBX_createHandlersToResp
00000000000096d0 T RDBX_createReaderFile
00000000000095a0 T RDBX_createReaderFileDesc
0000000000003530 T RDBX_createRespToFileWriter
0000000000005c60 T RDBX_createRespToRedisFd
0000000000006380 T RDBX_createRespToRedisTcp
00000000000059d0 T RDBX_enumerateCmds
0000000000005c20 T RDBX_setPipelineDepth
0000000000005a00 T RDBX_writeFromCmdNumber
```


api/librdb-ext-api.h
```diff
@@ -69,6 +73,7 @@ typedef struct RdbxToJsonConf {
     RdbxToJsonEnc encoding;  /* Encoding format for the resulting JSON */
     int includeAuxField;     /* Set to include auxiliary fields in JSON output */
     int includeFunc;         /* Set to include functions in JSON output */
+    int includeStreamMeta;   /* Set to include Stream meatadata in JSON output */
```
tezc, line 76:
```suggestion
    int includeStreamMeta;   /* Set to include Stream metadata in JSON output */
```

Makefile
```diff
@@ -30,6 +31,11 @@ distclean: clean
 example: all
 	cd examples && export LD_LIBRARY_PATH=../lib && ./example1
 
+# ------------------------- DEBUG -------------------------------------
+
+noopt:
+	OPTIMIZATION="-O0" LIBRDB_DEBUG=1 $(MAKE)
```
tezc, line 37: :rocket: I guess we can also pass sanitizer flags like that.

deps/redis/rax_malloc.h
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * This file is used in order to change the Rax allocator at compile time.
+ * Just define the following defines to what you want to use. Also add
+ * the include of your alternate allocator if needed (not needed in order
+ * to use the default libc allocator). */
+
+#ifndef RAX_ALLOC_H
+#define RAX_ALLOC_H
+//#include "zmalloc.h"
+#include "malloc.h"
```
tezc, line 41: `malloc.h` is a deprecated header, it might not exist on some systems. We are supposed to use `#include <stdlib.h>` for malloc.

src/cli/rdb-cli.c
```diff
@@ -372,9 +377,13 @@ int main(int argc, char **argv)
...
     }
 
-    if ((logfile = fopen(options.logfilePath, "w")) == NULL) {
-        printf("Error opening log file for writing: %s \n", options.logfilePath);
-        return RDB_ERR_GENERAL;
+    if (strcmp(options.logfilePath, "-") == 0) {
+        logfile = stdout;
+    } else {
+        if ((logfile = fopen(options.logfilePath, "w")) == NULL) {
+            printf("Error opening log file for writing: %s \n", options.logfilePath);
```
tezc, line 384: We can also print failure reason with `strerror(errno)`.

src/ext/handlersToJson.c
```diff
@@ -414,6 +443,147 @@ static RdbRes toJsonFunction(RdbParser *p, void *userData, RdbBulk func) {
...
+        ctx->state = R2J_IN_STREAM_ENTRIES_PAIRS;
+    } else {
+        fprintf(ctx->outfile, "} }");
+        ctx->state = R2J_IN_STREAM_ENTRIES;
+    }
+    return RDB_OK;
+}
+
+RdbRes toJsonStreamMetadata(RdbParser *p, void *userData, RdbStreamMeta *meta) {
+    UNUSED(p);
```
tezc, line 483:
```suggestion
```

seems like not unused.

src/ext/handlersToJson.c
```diff
@@ -414,6 +443,147 @@ static RdbRes toJsonFunction(RdbParser *p, void *userData, RdbBulk func) {
...
+    fprintf(ctx->outfile, "],\n      \"length\": %lu, ", meta->length);
+    fprintf(ctx->outfile, "\n      \"entriesAdded\": %lu, ", meta->entriesAdded);
+    fprintf(ctx->outfile, "\n      \"firstID\": \"%lu-%lu\", ", meta->firstID.ms, meta->firstID.seq);
+    fprintf(ctx->outfile, "\n      \"lastID\": \"%lu-%lu\", ", meta->lastID.ms, meta->lastID.seq);
+    fprintf(ctx->outfile, "\n      \"maxDelEntryID\": \"%lu-%lu\",", meta->maxDelEntryID.ms, meta->maxDelEntryID.seq);
+    return RDB_OK;
+}
+
+RdbRes toJsonStreamNewCGroup(RdbParser *p, void *userData, RdbBulk grpName, RdbStreamGroupMeta *meta) {
+    UNUSED(p);
```
tezc, line 500:
```suggestion
```
seems like not unused.

src/ext/handlersToJson.c
```diff
@@ -414,6 +443,147 @@ static RdbRes toJsonFunction(RdbParser *p, void *userData, RdbBulk func) {
...
+    }
+    fprintf(ctx->outfile, "%s        {\"name\": \"%s\", \"lastid\": \"%lu-%lu\", \"entriesRead\": %lu",
+            prefix, grpName, meta->lastId.ms, meta->lastId.seq, meta->entriesRead);
+
+    ctx->state = R2J_IN_STREAM_CG;
+    return RDB_OK;
+}
+
+RdbRes toJsonStreamCGroupPendingEntry(RdbParser *p, void *userData, RdbStreamPendingEntry *pe) {
+    UNUSED(p);
```
tezc, line 526:
```suggestion
```

src/ext/handlersToJson.c
```diff
@@ -414,6 +443,147 @@ static RdbRes toJsonFunction(RdbParser *p, void *userData, RdbBulk func) {
...
+                        "toJsonStreamCGroupPendingEntry(): Invalid state value: %d", ctx->state);
+        return (RdbRes) RDBX_ERR_R2J_INVALID_STATE;
+    }
+    fprintf(ctx->outfile, "%s\n           { \"sent\": %lu, \"id\":\"%lu-%lu\", \"count\": %lu }",
+            prefix, pe->deliveryTime, pe->id.ms, pe->id.seq, pe->deliveryCount);
+    return RDB_OK;
+}
+
+RdbRes toJsonStreamNewConsumer(RdbParser *p, void *userData, RdbBulk consName, RdbStreamConsumerMeta *meta) {
+    UNUSED(p);
```
tezc, line 545:
```suggestion
```

src/ext/handlersToResp.c
```diff
@@ -89,9 +125,17 @@ static void deleteRdbToRespCtx(RdbParser *p, void *context) {
...
+
+    if(ctx->streamCtx.groupPel)
+        deletePendingEntriesList(p, &ctx->streamCtx.groupPel);
+
     if (ctx->keyCtx.key != NULL)
         RDB_bulkCopyFree(p, ctx->keyCtx.key);
-    ctx->keyCtx.key = NULL;
+
+    if(ctx->streamCtx.consName)
+        RDB_bulkCopyFree(p, ctx->streamCtx.consName);
```
tezc, line 138: Maybe you can consider adding null check into `RDB_bulkCopyFree()`. So, you won't have to check it on each call and it would be in line with stdlib `free()` as we can call it null ptr. We can also provide same behavior for other `free()` like functions in the API.

src/ext/respToRedisLoader.c
```diff
@@ -65,8 +65,10 @@ static void onReadRepliesError(RdbxRespToRedisLoader *ctx) {
     }
 }
 
-/* Read 'numToRead' replies from the socket.
- * Return 0 for success, 1 otherwise. */
+/* Read 'numToRead' replies from the socket. * Return 0 for success, 1 otherwise.
+ *
+ * TODO: Support nonblocking configuration. If EAGAIN, propaagate it to parser's caller
```
tezc, line 70:
```suggestion
 * TODO: Support nonblocking configuration. If EAGAIN, propagate it to parser's caller
```

src/ext/respToRedisLoader.c
```diff
@@ -65,8 +65,10 @@ static void onReadRepliesError(RdbxRespToRedisLoader *ctx) {
...
+ *       and let it to decide whether it wants to retry now or later or change to blocking mode */
```
tezc, line 71: Isn't it blocking now? Where do you set it to nonblocking? I couldn't find it.

src/lib/parser.c
```diff
@@ -1171,6 +1175,42 @@ static RdbStatus zsetZiplistItem(RdbParser *p, BulkInfo *ziplistBulk) {
     return RDB_STATUS_OK;
 }
 
+static RdbStatus unzipStreamListpack(RdbParser *p, char *nodekey, unsigned char *lp) {
+    UNUSED(nodekey);
```
tezc, line 1179:
```suggestion
```
leftover?

src/lib/parser.c
```diff
@@ -1171,6 +1175,42 @@ static RdbStatus zsetZiplistItem(RdbParser *p, BulkInfo *ziplistBulk) {
...
+            if (!allocEmbeddedBulk(p, field, field_len, 0, &embField))
+                return RDB_STATUS_ERROR;
+
+            if (!allocEmbeddedBulk(p, value, value_len, 0, &embValue))
+                return RDB_STATUS_ERROR;
+
+            registerAppBulkForNextCb(p, &embField.binfo);
+            registerAppBulkForNextCb(p, &embValue.binfo);
+            CALL_HANDLERS_CB(p,
+                             restoreEmbeddedBulk(p, &embField);restoreEmbeddedBulk(p, &embValue);,
```
tezc, line 1202:
```suggestion
                             restoreEmbeddedBulk(p, &embField);restoreEmbeddedBulk(p, &embValue),
```
:cold_sweat: Macros are bit hard to read and debug, took me a while to understand the syntax. Last `;` is extra here.

src/lib/parser.c
```diff
@@ -2013,8 +2070,272 @@ RdbStatus elementModule(RdbParser *p) {
...
+RdbStatus rdbLoadMillisecTime(RdbParser *p, int64_t *val) {
+    BulkInfo *binfo;
+    IF_NOT_OK_RETURN(rdbLoad(p, 8, RQ_ALLOC, NULL, &binfo));
+    /*** ENTER SAFE STATE ***/
+    *val = *((int64_t *) binfo->ref);
+    if (p->rdbversion >= 9) /* see comment above */
+        memrev64ifbe(val);
+    return RDB_STATUS_OK;
+}
+
```
tezc, line 2338: Not suggesting a change here but endianness checks in redis reminds me this: https://commandcenter.blogspot.com/2012/04/byte-order-fallacy.html

In summary, instead of doing `memcpy/load + reverse`, if you do `byte shift + or`, you don't need to check endianness at all.

src/lib/parser.h
```diff
@@ -73,6 +73,16 @@ static inline void unused(void *dummy, ...) { (void)(dummy);}
...
 
+/* To save cycles, some places in code makes direct state-transition without
+ * passing through the parser main loop. When switch compilation LIBRDB_DEBUG is
+ * set, DBG_RETURN() enforces return to main loop, and in turn, to print the
+ * state transition */
+#if (LIBRDB_DEBUG)
+    #define DBG_RETURN(cmd) return (cmd)
+#else
+    #define DBG_RETURN(cmd) cmd
+#endif
```
tezc, line 84: I think this is going to hurt readability for the new folks (and for me :smile: ) . It's really difficult to consider both cases: if compiled with a flag it will return, otherwise it will continue. 

Maybe this is not the bottleneck at all, did you benchmark it? Maybe we can go with `return` until we prove it is the bottleneck.

src/lib/parserRaw.c
```diff
@@ -784,6 +784,232 @@ RdbStatus elementRawModule(RdbParser *p) {
...
+        ST_LOAD_NEXT_LP_STR1_RETURN,       /*   Complete loading node-key of next LP                                   */
+        ST_LOAD_NEXT_LP_STR2_RETURN,       /*   Complete loading next LP                                               */
+        ST_LOAD_METADATA,                  /* Load Stream metadata                                                     */
+        ST_LOAD_NEXT_CG_IS_MORE,           /* While more CG (Consumer groups) to load                                  */
+        ST_LOAD_NEXT_CG_STR_RETURN,        /*   Complete loading CG name                                               */
+        ST_LOAD_GLOBAL_PEL,                /*   Load CG's global PEL (Pending Entries List)                            */
+        ST_LOAD_NUM_CONSUMERS,             /*   Load number of consumers of current CG                                 */
+        ST_LOAD_NEXT_CONSUMER,             /*     Load next consumer                                                   */
+        ST_LOAD_NEXT_CONSUMER_STR_RETURN,  /*     Complete loading consumer name. Load consumder PEL.                  */
```
tezc, line 800:
```suggestion
        ST_LOAD_NEXT_CONSUMER_STR_RETURN,  /*     Complete loading consumer name. Load consumer PEL.                  */
```

src/lib/parserRaw.c
```diff
@@ -784,6 +784,232 @@ RdbStatus elementRawModule(RdbParser *p) {
...
+    } ;
+
+    ElementRawStreamCtx *streamCtx = &p->elmCtx.rawStream;
+    RawContext *rawCtx = &p->rawCtx;
+    UNUSED(p, rawCtx);
```
tezc, line 805:
```suggestion
```

src/cli/rdb-cli.c
```diff
@@ -372,9 +377,13 @@ int main(int argc, char **argv)
         return RDB_ERR_GENERAL;
     }
 
-    if ((logfile = fopen(options.logfilePath, "w")) == NULL) {
-        printf("Error opening log file for writing: %s \n", options.logfilePath);
-        return RDB_ERR_GENERAL;
+    if (strcmp(options.logfilePath, "-") == 0) {
```
tezc, line 380: Should user pass `-l -` ? Is this a convention? If not, it seemed a bit error-prone to me. Maybe `-l stdout` is better (or something else without using `-`).

src/lib/parser.h
oranagra: i don't like that at all.
i think flow control (return, break, if) are usually very dangerous to hide in macros.
making them optional seems much much worse.

oranagra (commented):
skimmed through it (skipped the tests as usual, other than a quick look at json and resp output).
took a few minor comments.

src/cli/rdb-cli.c
moticless: Good idea. thanks.

src/ext/handlersToResp.c
moticless: Right. thanks.

src/lib/parser.c
moticless: I followed redis implementation. Better to be aligned at this stage. If you think there is an option to enhance redis code, then please consider review it file [rdb.c](https://github.com/redis/redis/blob/unstable/src/rdb.c).

src/cli/rdb-cli.c
moticless: As Oran mentioned below, using `stdout` might be an even less favorable option. I will discard this option.

api/librdb-api.h
```diff
@@ -148,25 +149,27 @@ typedef struct RdbKeyInfo {
...
+    uint64_t length;           /* Current number of elements inside this stream. */
+    uint64_t entriesAdded;     /* All time count of elements added. */
+    RdbStreamID firstID;
+    RdbStreamID lastID;
+    RdbStreamID maxDelEntryID; /* maximum deleted entry id */
 } RdbStreamMeta;
 
 typedef struct RdbStreamPendingEntry {
-    long long deliveryTime;
+    RdbStreamID id;
```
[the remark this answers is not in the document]
moticless: The declaration of this callback is already too long (and i am too lazy). I prefer to leave it as is if you don't mind.

src/ext/readerResp.c
```diff
@@ -11,6 +11,8 @@ static inline void unused(void *dummy, ...) { (void)(dummy);}
 
 #define MAX_RSP_BULK_SIZE 1024*1024
 
+#define MAX_ARRAY_ELEMENTS ((1LL<<32) - 1)
```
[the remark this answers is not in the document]
moticless: We do. At hiredis, on [read.h](https://github.com/redis/hiredis/blob/869f3d0ef1513dd0258ad7190c9914df16dcc4a4/read.h#L70).

src/lib/parser.c
tezc: As I said, I'm not suggesting a change (too late for it). Just pointing another way for future reference.

src/lib/parser.h
moticless: The name of the macro actually place upfront exactly what it means which  is RETURN on DBG.

This is very helpfull to investigate corrupted RDB. Maybe the following detailed output that reflects each step transition  might convience you otherwise:

```
➜  push-to-librdb git:(add-stream-support) make noopt > /dev/null                        
➜  push-to-librdb git:(add-stream-support) ./bin/rdb-cli ./test/dumps/stream_v11.rdb resp > /dev/null
➜  push-to-librdb git:(add-stream-support) cat ./rdb-cli.log 
INFO  : RDBX_createReaderFile: Initialized with file ./test/dumps/stream_v11.rdb
INFO  : Finalizing parser configuration
INFO  : Start processing RDB source
INFO  : The parsed RDB file version is: 11
WARN  : Target Redis version is not configured! Set it to Redis version: 7.2
INFO  : rdbversion=11
DEBUG : [0x00000000] elementRdbHeader(State=0)
DEBUG : [0x00000009] elementNextRdbType(State=0) [Opcode=250]
DEBUG : [0x0000000a] elementAuxField(State=0)
DEBUG : [0x0000001c] elementNextRdbType(State=0) [Opcode=250]
DEBUG : [0x0000001d] elementAuxField(State=0)
DEBUG : [0x0000002a] elementNextRdbType(State=0) [Opcode=250]
DEBUG : [0x0000002b] elementAuxField(State=0)
DEBUG : [0x00000036] elementNextRdbType(State=0) [Opcode=250]
DEBUG : [0x00000037] elementAuxField(State=0)
DEBUG : [0x00000045] elementNextRdbType(State=0) [Opcode=250]
DEBUG : [0x00000046] elementAuxField(State=0)
DEBUG : [0x00000051] elementNextRdbType(State=0) [Opcode=254]
DEBUG : [0x00000052] elementSelectDb(State=0) [SelectDB=0]
DEBUG : [0x00000053] elementNextRdbType(State=0) [Opcode=251]
DEBUG : [0x00000054] elementResizeDb(State=0)
DEBUG : [0x00000056] elementNextRdbType(State=0) [Opcode=021]
DEBUG : [0x00000057] elementNewKey(State=0) [key=mystream]
DEBUG : [0x00000060] elementStreamLP(State=0) [key=mystream]
DEBUG : [0x00000061] elementStreamLP(State=1) [key=mystream]
DEBUG : [0x000000fa] elementStreamLP(State=1) [key=mystream]
DEBUG : [0x000000fa] elementStreamLP(State=2) [key=mystream]
DEBUG : [0x00000113] elementStreamLP(State=3) [key=mystream]
DEBUG : [0x00000126] elementStreamLP(State=4) [key=mystream]
DEBUG : [0x0000013f] elementStreamLP(State=4) [key=mystream]
DEBUG : [0x0000013f] elementStreamLP(State=5) [key=mystream]
DEBUG : [0x00000140] elementStreamLP(State=6) [key=mystream]
DEBUG : [0x0000015c] elementStreamLP(State=7) [key=mystream]
DEBUG : [0x0000015c] elementStreamLP(State=6) [key=mystream]
DEBUG : [0x00000178] elementStreamLP(State=7) [key=mystream]
DEBUG : [0x00000188] elementStreamLP(State=7) [key=mystream]
DEBUG : [0x00000188] elementStreamLP(State=3) [key=mystream]
DEBUG : [0x0000019b] elementStreamLP(State=4) [key=mystream]
DEBUG : [0x000001b4] elementStreamLP(State=4) [key=mystream]
DEBUG : [0x000001b4] elementStreamLP(State=5) [key=mystream]
DEBUG : [0x000001b5] elementStreamLP(State=6) [key=mystream]
DEBUG : [0x000001d1] elementStreamLP(State=7) [key=mystream]
DEBUG : [0x000001e1] elementStreamLP(State=7) [key=mystream]
DEBUG : [0x000001e1] elementStreamLP(State=3) [key=mystream]
DEBUG : [0x000001e1] elementEndKey(State=0)
DEBUG : [0x000001e1] elementNextRdbType(State=0) [Opcode=255]
DEBUG : [0x000001e2] elementEndOfFile(State=0) >>> [Status=(ENDED)]
INFO  : Parser done
DEBUG : RDB Reader: Closing file ./test/dumps/stream_v11.rdb        
```

src/ext/handlersToResp.c
```diff
@@ -6,9 +6,15 @@
 #include "../../deps/redis/crc64.h"
 #include "../../deps/redis/util.h"
 #include "../../deps/redis/endianconv.h"
+#include "../../deps/redis/rax.h"
 
+#define DEF_TARGET_VERSION_STRING "5.0"
```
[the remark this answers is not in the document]
moticless: Forgot to handle it. Modified it to take the latest version if not configured and output `WARN  : Target Redis version is not configured! Set it to Redis version: 7.2`

src/lib/parser.h
tezc: I meant you can use `return` unconditionally. So, you can still get debug logs.
I see from the above comment, you added that macro for better performance. I'm just not sure if it really impacts the performance. 

Anyway, leaving decision to you guys.

moticless: I know that most probably it won't have any impact in isolated test. The reason I feel ambigious about whether to remove this optimization or not is because every step in any given state of the parser can implicitly return due to partial data (i.e. will return `RDB_STATUS_WAIT_MORE_DATA`) without any real issue. This behavior is an intentional design choice to support async parsing.

src/ext/readerResp.c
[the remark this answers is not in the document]
oranagra: ok.. redis doesn't have that limit.. maybe add a comment.

src/lib/parser.h
moticless: I prefer to drop this macro rather than complicates the readability of this code even more (We still goanna have printouts excepts of the transitions that are being optimized.

src/ext/handlersToResp.c
```diff
@@ -260,13 +275,70 @@ static inline RdbRes writevWrap(RdbxToResp *ctx, struct iovec *iov, int cnt, int
...
+        if (res == RDB_OK)
+            RDB_reportError(ctx->parser, RDB_ERR_GENERAL, "Failed to writev() RESP");
+
         return RDB_getErrorCode(ctx->parser);
     }
 
     return RDB_OK;
 }
 
+static inline RdbRes sendFirstRawFrag(RdbxToResp *ctx, RdbBulk frag, size_t fragLen) {
```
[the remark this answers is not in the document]
oranagra: btw, why did you move it?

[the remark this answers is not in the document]
moticless: The file is arranged to the following sections: 
1. Mixture of static funcs
2. Handling common callbacks of type toRespXXX
3. Handling data callbacks of type toRespXXX
4. Handling raw callbacks of type toRespXXX
5. LIB API funcs

Moved it to first section.

deps/redis/librdb-hidden-api.h
```diff
@@ -0,0 +1,27 @@
...
+ * specific symbols from librdb.so that are needed by librdb-ext.so. These symbols
+ * won't be documented in the API, but they will be available for use. Since we
+ * compile by default with hidden visibility, we need to mark them explicitly as
+ * visible using the following macro.
+ *
+ * While it might initially seem cumbersome to manually designate each function as
+ * visible, this approach has its benefits. By carefully selecting the functions
+ * to expose, we are encouraged to contemplate what we are exposing and why.
+ * Moreover, it helps prevent the cluttering clients of librdb with unnecessary
+ * symbols.
```
oranagra, line 18: @yossigo maybe you can think of a better idea or have a strong opinion here.
it means that whoever links with our library gets many (unprefixed) APIs exposed, and possibly colliding.
e.g. raxNext, d2string, etc.

moticless: Another simple and effective option is to prefix the few functions with something like `RDB_HIDDEN_`.

src/lib/parser.c
```diff
@@ -1171,6 +1179,42 @@ static RdbStatus zsetZiplistItem(RdbParser *p, BulkInfo *ziplistBulk) {
...
+            if (!allocEmbeddedBulk(p, field, field_len, 0, &embField))
+                return RDB_STATUS_ERROR;
+
+            if (!allocEmbeddedBulk(p, value, value_len, 0, &embValue))
+                return RDB_STATUS_ERROR;
+
+            registerAppBulkForNextCb(p, &embField.binfo);
+            registerAppBulkForNextCb(p, &embValue.binfo);
+            CALL_HANDLERS_CB(p,
+                             restoreEmbeddedBulk(p, &embField);restoreEmbeddedBulk(p, &embValue),
```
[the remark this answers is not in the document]
moticless: All this line is the `finalize` operation. I agree. it is ugly. What about:
```
            CALL_HANDLERS_CB(p,
                             { /*finalize either on success or failure*/
                                 restoreEmbeddedBulk(p, &embField);
                                 restoreEmbeddedBulk(p, &embValue);
                             },
                             RDB_LEVEL_DATA,
                             rdbData.handleStreamItem,
                             &id,
                             embField.binfo.ref,
                             embValue.binfo.ref,
                             numfields);
```

deps/redis/librdb-hidden-api.h
tezc: Is there a specific reason for making extension api a separate library? I mean, we can generate a single so file and still provide two headers. Users who want to use extension API can include extension header and use those extra functions. (As as bonus, maintaining single file is a lot easier than maintaining two files)

moticless: We started with the easy choice of single library, but then raised aspects of modularity, seperation of concerns, maybe versioning, etc. And there is also my internal thought of having it in the far future as part of redis.

src/ext/handlersToResp.c
```diff
@@ -25,6 +25,7 @@ typedef struct RedisToRdbVersion {
 } RedisToRdbVersion;
 
 const RedisToRdbVersion redisToRdbVersion[] = {
+        {"7.2", VER_VAL(7,2), 12}, // TODO: Truly support 12 and update released redis version.
```
oranagra: i fear it is all too easy to forget about these. both the opcodes we already have in unstable, and any that we'll have later in 8.0 (we won't have the rdb version bump as a trigger to handle them).

i suppose we can split such opcodes into two categories:
1. new opcodes for existing types, i.e. all of the sudden hash keys are encoded differently, and all tests will fail.
2. new opcodes that are used only in a specific case (like the cluster ones added in unstable).

maybe we can think of some way to have a test that greps all the opcodes in the redis source code and validates they're all supported in librdb? i.e. just match the opcodes / numbers, doesn't test their functionality.

anyway, that's just some thought for the future. not a blocker for this PR

moticless: I already pushed a [commit](https://github.com/redis/librdb/pull/22/commits/ee0f60d2fd2fa2f89b9f332ed009c17407626403) that support the new opcode. I also modified the line to be  {"99.99", VER_VAL(99,99), 12}, to highlight that this value is not yet set. It doesn't have any side effect, just for the completeness.

moticless: Lookup for function `elementSlotInfo()`. `99.99` tells (at least to me) that it is WIP.
I will consider how we can memcmp and alert on changes in opcodes by reading unstable/rdb.h.

deps/redis/librdb-hidden-api.h
moticless: To workaround the symbols issues and API exposure I already modified the code to have in the core librdb API:
```
typedef struct RdbNonDocApi RdbNonDocApi; /* incomplete struct */
_LIBRDB_API RdbNonDocApi *RDB_getNonDocApi();
```
If you don't mind maybe let's leave it as is, until it will meet us again. Or do it later with lower priority.

oranagra: that's being creative :)
another simple option is to just include these files (e.g. rax.c) in both projects, if the symbols are hidden, won't cause any issues other than a few more kb in the binaries.

oranagra: i'm assuming that they don't interact.
i.e. that we don't pass a rax pointer created by one lib, to be accessed by the other.
also, make sure that other than `visibility=hidden` you also have `-Bsymolic`

moticless: They don't. 

Why to use `-Bsymbolic` and defer symbol resolution until runtime?

oranagra: AFAIK what it means is that local symbols are resolved within the library and won't attempt to be linked outside the library at runtime.
i.e. if your library uses raxSeek, even if its visibility is hidden, and external libraries won't be able to use it, there's still a chance your library will use a raxSeek provided by another library, instead of the local one

maybe i'm wrong. do you have other info?

moticless: I misunderstood this flag based on obsolete documentation 
> This option is somewhat historic, in that it was primarily designed for use in building the runtime linker itself.

Now I read that 
> Alert linker to search this shared object before the executable for symbols

But all the internal functions got resolved and doesn't appear in the symbols of the shared object. Verified by `nm -D`. And if it weren't resolved ... can this flag might impact someone that is integrating our libraries.

oranagra: a library can use `malloc` and `free` and then when it is linked with an executable, it can override these and provide an alternative implementation.
i.e. even if both your libraries don't expose `raxNext` and they have their own implementation, they may end up using a `raxNext` that's provided by the executable.
the great risk is that in this case, they're both different versions of redis's rax, but with some small differences, and one can end up using inline functions from one rax.h and non-inline functions from another, and if some struct has a different layout it can be catastrophic. (based on real life experience with sds.c)

moticless: >  even if both your libraries don't expose raxNext and they have their own implementation, they may end up using a raxNext that's provided by the executable.

I don't think it is true. If raxNext() doesn't appear in the dynamic symbol table (output of nm -D) of the library, it suggests that the symbol is not exposed for dynamic linking, which means it cannot be overridden in the usual dynamic linking manner. And therefore we don't have a problem and `-Bsymbolic` can be discarded.

src/ext/handlersToResp.c
```diff
@@ -930,6 +907,12 @@ static RdbRes toRespRestoreFragEnd(RdbParser *p, void *userData) {
 
 _LIBRDB_API RdbxToResp *RDBX_createHandlersToResp(RdbParser *p, RdbxToRespConf *conf) {
     RdbxToResp *ctx;
+    static int crcInitalized = 0;
+
+    if (!crcInitalized) {
+        ++crcInitalized;
+        crc64_init();
+    }
```
oranagra, line 915: isn't this dangerous when the lib is used from multiple threads (like we discussed in the past)?

moticless: To my best understanding, it is generally safe for two different threads to write the same value to a variable (int this case it is a table) as long as there are no race conditions or other synchronization issues involved. Writing the same value(s) to a variable simultaneously won't cause data corruption or unexpected behavior by itself. 

If you still prefer to synchronize it, I can modify it to :

```
    static int crcInitalized = 0;
    static pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;
    if (unlikely(!crcInitalized)) {
        /* If not init, acquire a lock to prevent concurrent init */
        pthread_mutex_lock(&mutex);
        if (!crcInitalized) {
            crc64_init();
            crcInitalized = 1;
       }
       pthread_mutex_unlock(&mutex);
    }
```

oranagra: It's probably safe to write to a variable but it could be that one thread concluded it can continue and the other thread re-init the table.
I'm afk and I don't know if re-init the table starts with zeroing or just overwrite, if it starts with zeroing your suggested solution isn't enough (thr mutex should wrap the initial flag check too

oranagra: Ohh I see the commit you made solves that

oranagra (approved):

src/ext/handlersToResp.c
moticless: You are right, there is a bug. 
I should have first call crc64_init() and only then set the flag and then claim my point :)
crc64_init doesn't zero the table. Just a set of assignments to the table.
In any way I pushed crc64_init_thread_safe() wrapper.

deps/redis/crc64.c
```diff
@@ -122,6 +122,20 @@ uint64_t crc64(uint64_t crc, const unsigned char *s, uint64_t l) {
     return crcspeed64native(crc64_table, crc, (void *) s, l);
 }
 
+/* Not sure if it is really required to use mutex here, but just in case */
```
tezc, line 125: I saw late but yes, we need a mutex here. Strictly speaking, there is still a race condition. For the outer if check, a thread might observe `crcInitalized = 1` even before `crc64_init()` is completed.

To solve the issue, we can get rid of the outer if check or make `crcInitalized` an atomic value. This is a known problem, please see https://en.wikipedia.org/wiki/Double-checked_locking

moticless: I am not sure if you followed the entire thread around it. 
There was an issue with my original implementation that set the flag before calling crc64_init(), but I still argue that:
> ... it is generally safe for two different threads to write the same value to a variable (int this case it is a table) as long as there are no race conditions or other synchronization issues involved. Writing the same value(s) to a variable simultaneously won't cause data corruption or unexpected behavior by itself.

In any way, to be on the safe I modified it to use mutex with double-check-locking for optimization.

moticless: I found an interesting article about [double-check-locking](https://www.aristeia.com/Papers/DDJ_Jul_Aug_2004_revised.pdf). Its conclusion is to use memory barriers. It is like falling down the rabbit hole.

I will go back to square one and claim that simply calling `crc64_init()` is lock-free programming withtout any risks.

> For example, if crc64_init() clears the table first or updates a table value twice, 

You can look at the code of `crc64_init`,  crc init doesn't memset the table and only makes write of constant values. And this not goanna change anytime soon.

The buttom line is maybe I should revert `crc64_init_thread_safe()` and call `crc64_init()` instead.

moticless: If there is one place I have the courage to write lock-free code, is during intialization alone and with function
that not goanna change. CRC tables are initialized this way for ages. For any other function I would have agree with you.

oranagra: FYI: https://github.com/redis/redis/pull/11988
i.e. i'm not certain things are always that simple and i do think it can change.

i rather not make assumptions, and either make a global init API with restrictions, or use safe locking without any tricks.

moticless: Yes, but he didn't touched `crc64_init()` :) Cool.

Ok, I got your point. I will create a PR with the following change:

```
void crc64_init_thread_safe(void) {
    static int crcInitalized = 0;
    static pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;

    pthread_mutex_lock(&mutex);
    if (!crcInitalized) {
        crc64_init();
        crcInitalized = 1;
    }
    pthread_mutex_unlock(&mutex);
}
```
It supposed to be fine even in case of reordering or multiprocessor machine. 

Thanks

PR #25, opened 2023-10-22
RESP2REDIS: Refine error reporting from server

moticless:
The api of RESP writer (which attached to RDB2RESP) is rather simple as it expects a plain iovec. This  API is not sufficient when it comes to play the RESP against live server such as the case of respToRedisLoader. In case of an error from the server it is desired to report an informative message to the user of the problematic **command** and **key**.  librdb does have mechanism to report on the problematic command even in case of pipeline of commands but it reports the entire command as a raw data with not distinction down to command and key. 

In order to keep the RESP writer API simple for the common case, the command and key will passed implicitly, hidden before iovec and the following struct reflects this layout. Respectively, respToRedisLoader will have the logic to cast iovec back to iovecExt.
```
typedef struct iovecExt {
    const char *cmd;
    const char *key;
    struct iovec iov[20];
} iovecExt;
```

### DIFF
--- a/src/ext/common.h
+++ b/src/ext/common.h
@@ -50,4 +50,20 @@ static inline void iov_plain(struct iovec *iov, const char *s, size_t l) {
     iov->iov_len = l;
 }
 
+/* The api of RESP writer is rather simple as it expects a plain iovec. This
+ * API is not sufficient when it comes to play the RESP against live server such
+ * as the case of respToRedisLoader. In case of an error from the server it is
+ * desired to report an informative message to the user of the problematic command
+ * and key.
+ *
+ * In order to keep the RESP writer API simple, the command and key will passed
+ * implicitly, hidden before iovec and the following struct reflects this layout.
+ * Respectively, respToRedisLoader will have the logic to cast iovec back to iovecExt.
+ */
+typedef struct iovecExt {
+    const char *cmd;
+    const char *key;
+    struct iovec iov[20];
+} iovecExt;
+
 #endif /*define RDBX_COMMON_H*/
--- a/src/ext/respToRedisLoader.c
+++ b/src/ext/respToRedisLoader.c
@@ -1,6 +1,7 @@
 #include <stdio.h>
 #include <string.h>
 #include <unistd.h>
+#include <stddef.h>
 #include <sys/socket.h>
 #include <netinet/in.h>
 #include <arpa/inet.h>
@@ -17,7 +18,7 @@
 #define PIPELINE_DEPTH_MAX        1000  /* limit the max value allowed to configure for pipeline depth */
 
 #define NUM_RECORDED_CMDS         400   /* Number of commands to backlog, in a cyclic array */
-#define RECORDED_DATA_MAX_LEN     40    /* Maximum payload size from any command to record into cyclic array */
+#define RECORDED_KEY_MAX_LEN     40    /* Maximum payload size from any command to record into cyclic array */
 
 #define REPLY_BUFF_SIZE           1024  /* reply buffer size */
 
@@ -29,7 +30,10 @@ struct RdbxRespToRedisLoader {
     struct {
         int num;
         int pipelineDepth;
-        char cmdPrefix[NUM_RECORDED_CMDS][RECORDED_DATA_MAX_LEN];
+        /* pointers to (static) strings that hold the template of the command sent (no char* allocation required) */
+        const char *cmd[NUM_RECORDED_CMDS];
+        /* strncpy() of the key sent */
+        char key[NUM_RECORDED_CMDS][RECORDED_KEY_MAX_LEN];
     } pendingCmds;
 
     RespReaderCtx respReader;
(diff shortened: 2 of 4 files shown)


src/ext/common.h
```diff
@@ -50,4 +50,20 @@ static inline void iov_plain(struct iovec *iov, const char *s, size_t l) {
     iov->iov_len = l;
 }
 
+/* The api of RESP writer is rather simple as it expects a plain iovec. This
+ * API is not sufficient when it comes to play the RESP against live server such
+ * as the case of respToRedisLoader. In case of an error from the server it is
+ * desired to report an informative message to the user of the problematic command
+ * and key.
+ *
+ * In order to keep the RESP writer API simple, the command and key will passed
```
tezc, line 59:
```suggestion
 * In order to keep the RESP writer API simple, the command and key will be passed
```

src/ext/respToRedisLoader.c
```diff
@@ -40,29 +44,13 @@ struct RdbxRespToRedisLoader {
...
-        RDB_reportError(ctx->p, (RdbRes) RDBX_ERR_RESP_WRITE,
-            "\nReceived Server error:\n\"%s\"\n\nGot failed on command [#%d] (First %d bytes):\n%s\n",
-            respReader->errorMsg,
-            ctx->respReader.countReplies,
-            RECORDED_DATA_MAX_LEN,
-            currCmdRecord);
-    }
+
+    RDB_reportError(ctx->p, (RdbRes) RDBX_ERR_RESP_WRITE,
+                    "\nerror from dst '-%s' on key '%s' on command '%s' (RESP Command #%lu)\n",
```
tezc, line 49:
```suggestion
                    "\nerror from dst '-%s' on key '%s' on command '%s' (RESP Command #%zu)\n",
```
Seems like this is size_t.

api/librdb-ext-api.h
```diff
@@ -147,12 +147,19 @@ _LIBRDB_API RdbxToResp *RDBX_createHandlersToResp(RdbParser *, RdbxToRespConf *)
...
+} RdbxRespWriterStartCmd;
+
 typedef struct RdbxRespWriter {
     void *ctx;
     void (*delete)(void *ctx);
 
     /* return 0 on success. Otherwise 1 */
-    int (*writev) (void *ctx, struct iovec *ioVec, int iovCnt, int startCmd, int endCmd);
+    int (*writev) (void *ctx, struct iovec *ioVec, int iovCnt,
+                   RdbxRespWriterStartCmd *startCmd, int endCmd);
```
tezc, line 162: maybe you can add a comment to explain startCmd and endCmd. It took me a while to understand it. e.g. endCmd is a boolean flag etc.

tezc (approved):
had a quick look, lgtm.

api/librdb-ext-api.h
```diff
@@ -139,20 +139,31 @@ _LIBRDB_API RdbxToResp *RDBX_createHandlersToResp(RdbParser *, RdbxToRespConf *)
...
 typedef struct RdbxRespWriter {
     void *ctx;
     void (*delete)(void *ctx);
 
     /* return 0 on success. Otherwise 1 */
-    int (*writev) (void *ctx, struct iovec *ioVec, int iovCnt, int startCmd, int endCmd);
+    int (*writev) (void *ctx,
+                   struct iovec *ioVec,              /* Standard C scatter/gather IO array */
+                   int iovCnt,                       /* Number of iovec elements */
+                   RdbxRespWriterStartCmd *startCmd, /* Not NULL if this is start of RESP command */
```
oranagra, line 164: maybe worth documenting the memory ownership model of startCmd

oranagra (changes requested):

src/ext/respToRedisLoader.c
```diff
@@ -323,16 +315,23 @@ _LIBRDB_API RdbxRespToRedisLoader *RDBX_createRespToRedisTcp(RdbParser *p,
...
     if (connect(sockfd, (struct sockaddr *) &server_addr, sizeof(server_addr)) == -1) {
         RDB_reportError(p, (RdbRes) RDBX_ERR_RESP2REDIS_INVALID_ADDRESS,
                         "Invalid tcp address (hostname=%s, port=%d)", hostname, port);
-        close(sockfd);
-        return NULL;
+        goto createErr;
     }
 
-    return RDBX_createRespToRedisFd(p, rdbToResp, auth, sockfd);
+    RdbxRespToRedisLoader *res = RDBX_createRespToRedisFd(p, rdbToResp, auth, sockfd);
```
tezc, line 327: I see inside RDBX_createRespToRedisFd(), it calls close() for fd in one error path. Probably, it is better to do it in a single place.

src/ext/respToRedisLoader.c
```diff
@@ -323,16 +315,23 @@ _LIBRDB_API RdbxRespToRedisLoader *RDBX_createRespToRedisTcp(RdbParser *p,
...
-        return NULL;
+        goto createErr;
     }
 
     if (connect(sockfd, (struct sockaddr *) &server_addr, sizeof(server_addr)) == -1) {
         RDB_reportError(p, (RdbRes) RDBX_ERR_RESP2REDIS_INVALID_ADDRESS,
                         "Invalid tcp address (hostname=%s, port=%d)", hostname, port);
-        close(sockfd);
-        return NULL;
+        goto createErr;
```
tezc, line 324: I think logging errno will be helpful

tezc (approved):
lgtm, just minor comments might be fixed in a later pr

src/ext/respToRedisLoader.c
moticless: right. thanks.

src/ext/respToRedisLoader.c
```diff
@@ -323,16 +315,23 @@ _LIBRDB_API RdbxRespToRedisLoader *RDBX_createRespToRedisTcp(RdbParser *p,
     if (inet_pton(AF_INET, hostname, &(server_addr.sin_addr)) <= 0) {
         RDB_reportError(p, (RdbRes) RDBX_ERR_RESP2REDIS_INVALID_ADDRESS,
                         "Invalid tcp address (hostname=%s, port=%d)", hostname, port);
-        close(sockfd);
-        return NULL;
+        goto createErr;
```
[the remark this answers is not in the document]
moticless: Note that the first error already `RDB_reportError()` which is the root cause. It doesn't really matter if close() got failed or not with errno.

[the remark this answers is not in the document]
tezc: not sure I understand but I see inet_pton does nto report much errors anyway. Probably not necessary here.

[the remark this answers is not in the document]
tezc: btw it will be a lot more useful for `connect()`.  Please consider adding to it down below.

tezc (approved):

PR #26, opened 2023-10-29
Add soname, pkgconfig and fix makefile issues

tezc:
- Renamed `VERSION` variable to `LIBRDB_VERSION`. 
  If outer makefile sets `VERSION` variable for some purpose, it overrides this one and causes a mess. 

- Added option to install static or dynamic libraries.
- Added basic pkgconfig support
- Added `soname` to the shared libraries. Some systems (e.g. redhat based ones) fetch library name from it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,8 +6,19 @@ INSTALL = /usr/bin/install -c
 BINDIR=$(DESTDIR)$(PREFIX)/bin
 LIBDIR=$(DESTDIR)$(PREFIX)/lib
 INCDIR=$(DESTDIR)$(PREFIX)/include/librdb/
+LIBRDB_INSTALL_SHARED:=yes
+LIBRDB_INSTALL_STATIC:=yes
 
-VERSION = $(shell grep -oP '(?<=LIBRDB_VERSION_STRING ")[0-9]+\.[0-9]+\.[0-9]+' ./src/lib/version.h)
+UNAME := $(shell uname)
+
+ifneq (,$(filter $(UNAME),OpenBSD FreeBSD NetBSD))
+	PKGCONFIGDIR = $(DESTDIR)$(PREFIX)/libdata/pkgconfig
+else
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -6,8 +6,19 @@ INSTALL = /usr/bin/install -c
 BINDIR=$(DESTDIR)$(PREFIX)/bin
 LIBDIR=$(DESTDIR)$(PREFIX)/lib
 INCDIR=$(DESTDIR)$(PREFIX)/include/librdb/
+LIBRDB_INSTALL_SHARED:=yes
+LIBRDB_INSTALL_STATIC:=yes
```
moticless, line 10: Where is this flag being modified? If not, then consider removing it.

tezc: This is the option to install libraries selectively. User is supposed to pass it like `make install LIBRDB_INSTALL_STATIC=no`. 
For example, Redhat complains about unused static libraries, I thought we can call make install with this option.

moticless: In that case i it should be internal logic of librdb `make install`, resolve platform and decide  whether to copy or not static libraries.

tezc: I didn't mean it is not possible to use static library on redhat. It is possible.

In our specific case, on redhat, rpm build process complains that `.a` libraries are installed but not part of the package. We are not going to use `.a` libs and I don't want to add them to the package. So, how do I solve this problem? Either I will not install them from the beginning or I will install them and then delete before starting this rpm generation process. 

Why not give this option to user not to install specific type of library? Some other user might even decide to use static library only and skip installing dynamic ones. 

Some libraries provide this option in their build system, e.g. [jemalloc](https://github.com/jemalloc/jemalloc/blob/e4817c8d89a2a413e835c4adeab5c5c4412f9235/Makefile.in#L568)
or if a library uses cmake, you can only build one kind of library at a time (configurable via parameter).

moticless: Got it. Thanks.

moticless (approved):

PR #27, opened 2023-11-01
RDB2RESP: Add single-db flag

moticless:
On RDB2RESP, add `singleDb` flag. i.e., avoid using the SELECT command. 
In turn all keys and data in the RDB will be stored in the default DB (index 0).
This approach can be helpful when dealing with scenarios such as data partitioned 
into multiple DBs where you need to merge them into a single DB. Be cautious of 
potential key conflicts in such cases. 

Add also makefile dependency for CLI to rebuild on each static librdb modification.

### DIFF
--- a/api/librdb-ext-api.h
+++ b/api/librdb-ext-api.h
@@ -132,6 +132,17 @@ typedef struct RdbxToRespConf {
      * this functionality to Redis OSS in the near future. */
      int supportRestoreModuleAux;
 
+    /* Single DB mode. i.e., avoid using the SELECT command. In turn all keys and
+     * data in the RDB will be stored in the default DB (index 0).
+     *
+     * This approach can be helpful when dealing with scenarios such as data
+     * partitioned into multiple DBs where you need to merge them into a single DB.
+     * Be cautious of potential key conflicts in such cases.
+     *
+     * Additionally, this can be useful when migrating data to Redis Enterprise,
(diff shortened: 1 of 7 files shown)


api/librdb-ext-api.h
```diff
@@ -132,6 +132,17 @@ typedef struct RdbxToRespConf {
...
      int supportRestoreModuleAux;
 
+    /* Single DB mode. i.e., avoid using the SELECT command. In turn all keys and
+     * data in the RDB will be stored in the default DB (index 0).
+     *
+     * This approach can be helpful when dealing with scenarios such as data
+     * partitioned into multiple DBs where you need to merge them into a single DB.
+     * Be cautious of potential key conflicts in such cases.
+     *
+     * Additionally, this can be useful when migrating data to Redis Enterprise,
```
tezc, line 142: shouldn't we avoid mentioning private implementation details here? It is not relevant to others and if that behavior changes one day, we'll need to update here too etc..

tezc (approved):

api/librdb-ext-api.h
oranagra: OSS cluster also doesn't support multiple dbs.
i think the first paragraph is enough.

oranagra (approved):

api/librdb-ext-api.h
moticless: cool. thanks.

tezc (approved):

PR #28, opened 2023-11-02
Fix module-Aux handling

moticless:
- parserRaw.c didn't init appropriately aggregation of bulks for module-aux. Refactor and call to `aggReset()`.
- module-aux could wrongly include REPLACE, ABSTTL, IDLETIME or FREQ flags.

### DIFF
--- a/src/lib/parserRaw.c
+++ b/src/lib/parserRaw.c
@@ -678,7 +678,7 @@ RdbStatus elementRawModule(RdbParser *p) {
                     updateElementState(p, ST_NEXT_OPCODE, 0);
                     break;
                 } else {
-                    assert(p->currOpcode == RDB_TYPE_MODULE_AUX);
+                    assert(p->currOpcode == RDB_OPCODE_MODULE_AUX);
 
                     /* Init Aggregator of bulks here since no new-key precedes module aux */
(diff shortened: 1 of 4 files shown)


src/lib/parserRaw.c
```diff
@@ -680,9 +679,9 @@ RdbStatus elementRawModule(RdbParser *p) {
                     break;
                 }
 
-                /* No new-key precedes module aux. Init Aggregator of bulks here.
-                 * Note that the call is made from a safe state */
-                aggAllocFirstBulk(p);
+                /* No new-key precedes module aux. Init Aggregator of bulks here. */
+                aggReset(p);
+
                 updateElementState(p, ST_AUX_START, 0);
```
oranagra, line 685: personally i think it's odd that this code is just placed outside the `RDB_TYPE_MODULE_2` condition.
i think it should be an `else` with separate scope, and a comment stating that it belongs to `RDB_OPCODE_MODULE_AUX`

oranagra (approved):

src/lib/parserRaw.c
moticless: Done. Thanks.

oranagra (approved):

PR #32, opened 2023-11-26
RDB2RESP: Fix functions registration

moticless:
In case of multiple levels registration of callbacks, RDB2RESP need to take
care to enforce that functions will be handled only by data-level (To simplify
implementation we avoid using FUNCTION RESTORE command and Instead, 
always use FUNCTION LOAD command. It is expected that there won't be 
many functions in the RDB file in which requires optimization).

### DIFF
--- a/api/librdb-ext-api.h
+++ b/api/librdb-ext-api.h
@@ -113,6 +113,12 @@ typedef struct RdbxToRespConf {
      * RESTORE command. */
     int delKeyBeforeWrite;
 
+    /* funcLibReplaceIfExist - If function-library with the same name is already
+     * exist in the target redis, then replace it rather than return failure.
+     * (Implemented in RESP by adding `REPLACE` flag to the `FUNCTION LOAD`
+     * command) */
+    int funcLibReplaceIfExist;
+
     /* If supportRestore, then data-types will be translated to RESTORE with
      * raw data instead of data-types commands. This is a performance optimization
      * that requires to be version aligned. */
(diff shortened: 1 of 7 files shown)


src/ext/handlersToResp.c
```diff
@@ -204,6 +204,9 @@ static RdbRes resolveSupportRestore(RdbParser *p, RdbxToResp *ctx) {
     /* Enforce RESTORE for modules. librdb cannot really parse high-level module object */
     RDB_handleByLevel(p, RDB_DATA_TYPE_MODULE, RDB_LEVEL_RAW);
 
+    /* Avoid RESTORE for functions. Redis doesn't have API to RESTORE functions */
```
oranagra, line 207: that's not true.
redis does have FUNCTION DUMP and FUNCTION RESTORE.
it has the same format as the RDB file, i.e. dumping dumping all functions at once.
so in theory instead of sending multiple FUNCTION LOAD commands, you can buffer that data and send s single FUNCTION RESTORE command.
not sure what's the benefit of that (maybe if there are a ton of functions (libraries), which is unlikely IMHO).

Maybe the benefit of such, is that you can use the FLUSH argument, and there's no need for an additional FUNCTION FLUSH before starting.

which brings me to realize that where we use FLUSHALL, we should also use FUNCTION FLUSH, and where we use DEL or REPLACE, we should do the same for functions. please look into it.

moticless: Missed that one. Thanks. I will expose configuration flag to decide whether to REPLACE or not on FUNCTION LOAD.

api/librdb-ext-api.h
```diff
@@ -113,6 +113,12 @@ typedef struct RdbxToRespConf {
      * RESTORE command. */
     int delKeyBeforeWrite;
 
+    /* funcLibReplaceIfExist - If function-library with the same name is already
+     * exist in the target redis, then replace it rather than return failure.
+     * (Implemented in RESP by adding `REPLACE` flag to the `FUNCTION LOAD`
+     * command) */
+    int funcLibReplaceIfExist;
```
oranagra, line 120: i suppose the default is off, and i didn't see it being initialized, i assume we're using memset of calloc.

oranagra (approved):

PR #33, opened 2023-12-04
Fix filedesc +  Fix state transition of stream

moticless:
- Fix file descriptor reader.  Was written naively without checking what happen if not all the data was read or for EINTR.
- Fix wrong fall-through in the states of stream element.

I will add tests on another commit.

### DIFF
--- a/src/lib/parserRaw.c
+++ b/src/lib/parserRaw.c
@@ -845,7 +845,7 @@ RdbStatus elementRawStreamLP(RdbParser *p) {
                     return RDB_STATUS_ERROR;
                 }
             }
-            updateElementState(p, ST_LOAD_NEXT_LP_IS_MORE, 1); /* fall-thru */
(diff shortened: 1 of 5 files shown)


src/lib/parserRaw.c
```diff
@@ -845,7 +845,7 @@ RdbStatus elementRawStreamLP(RdbParser *p) {
                     return RDB_STATUS_ERROR;
                 }
             }
-            updateElementState(p, ST_LOAD_NEXT_LP_IS_MORE, 1); /* fall-thru */
+            return updateElementState(p, ST_LOAD_NEXT_LP_IS_MORE, 1); /* fall-thru */
```
tezc, line 848:
```suggestion
            return updateElementState(p, ST_LOAD_NEXT_LP_IS_MORE, 1);
```
No more fall-thru right?

tezc (approved):

oranagra (approved):

PR #36, opened 2023-12-10
open fd with fdopen() to improve performance

moticless:
Optimize reading from fd by using fdopen()


oranagra (commented):
i don't like this approach.
it feels like there are bugs (when part of the data is already buffered).

maybe you should use `fdopen` instead and let libc do the caching (assuming it does read-ahead, please confirm)

tezc:
I agree with @oranagra. If possible, you can use FILE api which handles caching. 

If this is really the bottleneck and if we want to implement caching ourselves, I guess you can take a look at implementations of `fread()`. I assume they use `readv()` for this purpose, just fyi.

moticless:
@oranagra , I shifted to use fdopen() (no escaped but to use _POSIX_C_SOURCE) and discarded input-buffer. Thanks.

oranagra:
did you benchmark the 3 alternatives?

moticless:
@oranagra , 
This change is based on some basic benchmarking that i made between plain read() vs. read()+input-buffer. 
fdopen() gives similar results to read()+input-buffer. Although the library was integrated into another utility (rl_rdbloader), the adjustment is relatively self-contained and provides a reliable indication of its effectiveness.

src/ext/readerFileDesc.c
```diff
@@ -74,19 +65,28 @@ RdbxReaderFileDesc *RDBX_createReaderFileDesc(RdbParser *p, int fd, int fdCloseW
...
         RDB_reportError(p, RDB_ERR_NONBLOCKING_FD,
-            "RDBX_createReaderFileDesc(): fd must be set to blocking mode");
+            "RDBX_createReaderFileDesc(fd=%d): fd must be set to blocking mode", fd);
+        return NULL;
+    }
+
+    FILE *file = fdopen(fd, "r");
+
+    if (file == NULL) {
+        RDB_reportError(p, RDB_ERR_NONBLOCKING_FD,
```
tezc, line 81: Is this error code accurate? This error is not actually about blocking/nonblocking fd. It looked a bit confusing to me. 
Btw, I'd also print `strerror(errno)` here to figure out why `fdopen()` failed.

src/ext/readerFileDesc.c
```diff
@@ -74,19 +65,28 @@ RdbxReaderFileDesc *RDBX_createReaderFileDesc(RdbParser *p, int fd, int fdCloseW
     if (flags==-1) {
 
         RDB_reportError(p, RDB_ERR_FAILED_GET_FD_FLAGS,
-            "RDBX_createReaderFileDesc(): Error getting file descriptor flags");
+            "RDBX_createReaderFileDesc(fd=%d): Error getting file descriptor flags", fd);
```
tezc, line 68: you may log `strerror(errno)` as well

oranagra (approved):
LGTM. no other comments.

tezc (approved):

api/librdb-ext-api.h
```diff
@@ -25,6 +25,9 @@ typedef enum {
     RDBX_ERR_READER_FILE_GENERAL_ERROR = _RDB_ERR_EXTENSION_FIRST,
     RDBX_ERR_RESP_FAILED_ALLOC,
 
+    /* reader errors */
+    RDBX_ERR_FAILED_FD_OPEN,
```
oranagra, line 29: considering that the API and maybe even ABI should be stable across versions, i'm not sure if we want the error to be that specific, and if being that specifc somehow serves the caller at all.
why not use RDBX_ERR_FAILED_OPEN_FILE?
i.e. the failure happened while preparing the file (descriptor) for reading.

src/ext/readerFileDesc.c
```diff
@@ -65,7 +65,7 @@ RdbxReaderFileDesc *RDBX_createReaderFileDesc(RdbParser *p, int fd, int fdCloseW
     if (flags==-1) {
 
         RDB_reportError(p, RDB_ERR_FAILED_GET_FD_FLAGS,
```
oranagra, line 67: same argument as above. seems too specific to me.

src/ext/readerFileDesc.c
```diff
@@ -78,7 +78,7 @@ RdbxReaderFileDesc *RDBX_createReaderFileDesc(RdbParser *p, int fd, int fdCloseW
     FILE *file = fdopen(fd, "r");
 
     if (file == NULL) {
-        RDB_reportError(p, RDB_ERR_NONBLOCKING_FD,
+        RDB_reportError(p, RDBX_ERR_FAILED_FD_OPEN,
```
oranagra: let's print errno or strerror

src/ext/respToFileWriter.c
```diff
@@ -54,7 +54,7 @@ RdbxRespToFileWriter *RDBX_createRespToFileWriter(RdbParser *p, RdbxToResp *rdbT
     } else {
         filePtr = fopen(filePath, "wb");
         if (filePtr == NULL) {
-            RDB_reportError(p, (RdbRes) RDBX_ERR_FAILED_OPEN_FILE, "createRespWriter: Failed to open file: %s",
+            RDB_reportError(p, RDB_ERR_FAILED_OPEN_FILE, "createRespWriter: Failed to open file: %s",
```
oranagra, line 57: errno is missing here too, and in the other calls for fopen. basically everywhere that we get a failure from a system / libc call

src/ext/respToFileWriter.c
```diff
@@ -20,7 +20,7 @@ static int respFileWritev(void *context, struct iovec *iov, int count,
     /* not optimized code */
     for (int i = 0 ; i < count ; ++i) {
         if (unlikely(fwrite(iov[i].iov_base, sizeof(char), iov[i].iov_len, ctx->filePtr) != iov[i].iov_len)) {
-            RDB_reportError(ctx->p, (RdbRes) RDBX_ERR_RESP_WRITE, "Failed to write RESP to file");
+            RDB_reportError(ctx->p, (RdbRes) RDBX_ERR_RESP_WRITE, "Failed to write RESP to file: %s (errno=%d)", errno);
```
tezc, line 23: I think parameter for `%s` is missing here. Btw, you can use `__attribute__ ((format (printf, 3, 4)))` here for this function so compiler can check mismatches.

tezc (approved):

PR #42, opened 2024-03-17
Aggregate reported-errors up to 2kb

moticless:


src/lib/parser.c
```diff
@@ -447,28 +448,41 @@ _LIBRDB_API RdbRes RDB_getErrorCode(RdbParser *p) {
...
     assert (e != RDB_OK && e != RDB_OK_DONT_PROPAGATE);
 
+    /* If error message is too long, then trim it in order to record this last message */
+    if (p->errorMsgAt > LAST_ERR_MSG_OFFSET) {
+        p->errorMsgAt = LAST_ERR_MSG_OFFSET;
+        p->errorMsgAt += snprintf(p->errorMsg + p->errorMsgAt,
+                                  MAX_ERROR_MSG - p->errorMsgAt,
+                                  "\n... last recorded error message: ...\n");
+    }
+    p->errorMsgAt += snprintf(p->errorMsg + p->errorMsgAt, MAX_ERROR_MSG - p->errorMsgAt, "[errcode=%d] ", e);
```
tezc, line 469: IIUC there is a chance that buffer can be full. So, this snprintf or the ones below can return amount of required space on failure. Then `p->errorMsgAt` can become larger than MAX_ERROR_MSG and cause some weird out of bounds issues? Maybe we should explicitly check snprintf/vsnprintf return values for errors?

moticless: I think the code is safe as long as `p->errorMsgAt` is correctly managed and we are using `snprintf` which protects from overflow. I also added overflow test. Do you still think there is still a risk?

tezc (approved):

PR #46, opened 2024-06-13
Integrate python json parser as a service for testing

moticless:
* One of the outputs of librdb is JSON. Until now there was a naive testing code that compared two json files, attempting to ignore JSON elements ordering. This PR replaces it with a python service, run in the background, read JSON filename from its STDIN, generates its signature by ignoring JSON elements order, and prints SHA256 signature to its STDOUT. test_common.c will fork this service. The reason it is running as a service is to avoid the overhead of starting a new process for each evaluation and the intensive use of it.

* In addition the optional json fields will be now in the form:
```
__aux__ : { …. }
__slotinfo__ : { …. }
__dbsize__ : { …. }
```

* In JSON, the value “-0.0” is a valid double, but we want to output it as 0. 

* Modified librdb tests to run with -O0 optimization instead of -O3


test/json_signature_generator.py
```diff
@@ -0,0 +1,83 @@
...
+
+            # Process the file
+            json_data = load_json(file_path)
+            sorted_json_data = sort_json(json_data)
+            sorted_json_string = json_to_string(sorted_json_data)
+            signature = generate_signature(sorted_json_string)
+
+            # Return the signature
+            print(signature)
+            logging.info(f"Signature generated for file: {file_path}")
```
tezc, line 73: This would normally break your protocol but it works as you disabled logging above right? 
but then, `logging.error()` lines won't print anything if you encounter a real problem?  Overall, it looks confusing to me. Maybe you can consider using `print(....)` API only.

moticless: Note that the logging configuration, which is in comment, is directed into a file:
```
# logging.basicConfig(filename='./json_signature_generator.log', level=logging.INFO,
#                     format='%(asctime)s - %(levelname)s - %(message)s')
```

test/test_common.c
```diff
@@ -551,65 +553,103 @@ static unsigned char xorstr(const char *str) {
...
+        dup2(pipe_out[1], STDOUT_FILENO); /* Redirect child STDOUT */
+        /* now replace child process with python service */
+        execlp("python3", "python3", "./test/json_signature_generator.py", NULL);
+        perror("execlp");
+        exit(1);
+    } else {
+        /*Parent*/
+        close(pipe_in[0]);  /* Close read end of pipe_in */
+        close(pipe_out[1]); /* Close write end of pipe_out */
+    }
```
tezc, line 590: Not sure about the logic here but please consider if there is a potential race condition here. e.g. parent exists and starts running tests while child is still in the above "if check"

test/json_signature_generator.py
tezc: A bit confusing to see `logging.info/error()` lines that are NOOP but it is okay ofcouse if you aware of it.

tezc (dismissed):

tezc:
LGTM. Just added minor comments.

test/test_common.c
moticless: The parent will block as long as it didn't read a single byte. I tested it by putting 10 seconds delay at the begning of the python code and the parent waited peacefully.

tezc (approved):

PR #47, opened 2024-06-13
Add support to hash-field-expiration (7.4)

moticless:
Required to support two new opcodes: 
`RDB_TYPE_HASH_METADATA` - While `RDB_TYPE_HASH` keep `key-value` pairs to rdb file, this one keep `key-value-expiry` tuples such that the value of `expiry` cannot be bigger than  2^48-1.

`RDB_TYPE_HASH_LISTPACK_EX` - While `RDB_TYPE_HASH_LISTPACK_EX` keep in listpack pairs of `key-value`, this one keep `key-value-expiry` tuples such that the value of `expiry` cannot be bigger than  2^48-1.


api/librdb-api.h
```diff
@@ -257,12 +257,14 @@ typedef struct RdbHandlersStructCallbacks {
     /* Callback to handle a listpack-based list value */
     RdbRes (*handleListLP)(RdbParser *p, void *userData, RdbBulk listpack);
 
-    /* Callback to handle a field-value pair within a plain-hash */
-    RdbRes (*handleHashPlain)(RdbParser *p, void *userData, RdbBulk field, RdbBulk value);
+    /* Callback to handle a field-value pair within a plain-hash. expireAt -1 if not set. */
```
tezc, line 260:
```suggestion
    /* Callback to handle a field-value pair within a plain-hash. expireAt is -1 if not set. */
```

tezc (approved):

tezc (approved):

PR #51, opened 2024-07-24
Support RDB_TYPE_HASH_METADATA relative time optimization

moticless:
Adapt librdb to modified layout of RDB_TYPE_HASH_METADATA which 
now store field's relative TTLs instead of absolute expiration time.


oranagra (approved):
please add some description

PR #53, opened 2024-08-08
RESP2REDIS: Should not fail trying to load empty module

moticless:
If RDB2RESP was configured to "supportRestoreModuleAux" and generates
RESTOREMODAUX commands, currently relevant only to Redis enterprise, 
then if RDB was generated by a server with some module, but user didn't 
make any use of that module, attempting to play it to another server that 
wasn't loaded with that module, the RDB parser will get fail. This is because 
the module always store something in the AUX field, and the RDB parser will try 
to load it.

In order to overcome this issue, A module that its AUX payload is less than
15 Bytes (including RDB version and checksum) counted as AUX field of an empty
Module (not in use), then the parser, when restoring the empty module, it
should ignore returned error: "-ERR Module X not found..."

Few changes were made to support it:
* Propagate `restoreSize` to RESP2REDIS handlers.
* ReaderResp gets now error-callback which can indicate to mask given error.
* RESP2REDIS indicates to ignore RESTOREMODAUX error message of type "-ERR Module  .. not found ..."
* Added tests accordingly.

Additionally added timeout to recv() from redis to avoid blocking forever. Irrelevant to this fix.


YaacovHazan (approved):

oranagra (approved):
reviewed by Yaacov

PR #66, opened 2025-01-30
Add option to hide keys in log and print sha256(key) instead

moticless:
* Extend librdb-api.h with `RDB_hideKeysInLog(RdbParser *p)` to hide keys in logs during errors by printing the first 8 hex digits of SHA256(key).
* Rename `./src/ext/common.h` to “`/src/ext/extCommon.h`
* Avoid printing twice the reported error message. (The first time occurs when calling to `RDB_reportError()` and the second time at the end of `parserMainLoop()`, by calling `printParserState()`). 
* On error, add to `printParserState()` also the value of `bytesRead` of the parser.
* On `elementEndKey()`, set `...dataType = RDB_DATA_TYPE_MAX;` for completeness.


tezc (approved):
glanced over changes quickly, it looks good to me.

PR #67, opened 2025-02-11
Add parsing support for RDB_OPCODE_RAM_LRU for Redis Ent (No-op)

moticless:
Trivial parsing for the opcode `RDB_OPCODE_RAM_LRU = 107`. Does nothing.
(Extend also rdb-cli to support `--ignore-checksum`)


oranagra (approved):
LGTM

PR #69, opened 2025-02-27
Fix recv() 120s Timeout on EAGAIN by Retrying Indefinitely

moticless:
Syncing with redis server may fail on configured 120-second timeout on recv()   
 in RESTORE MODE with a big key, say >4GB. It takes Redis server a long time to  
 process the RESTORE command before ACK. The solution is to modify the logic to  
 continuously retry without a timeout. An appropriate log message will be printed 
 every 10 seconds, such as: "No reply from Redis server for 10 seconds."         
                                                                                 
 If requested, we can expose conifguration to fine tune timeout duration and retries.


test/test_rdb_to_redis.c
```diff
@@ -580,6 +654,7 @@ int group_rdb_to_redis(void) {
             cmocka_unit_test_setup(test_rdb_to_redis_multiple_dbs, setupTest),
             cmocka_unit_test_setup(test_rdb_to_redis_function, setupTest),
             cmocka_unit_test_setup(test_rdb_to_redis_func_lib_replace_if_exist, setupTest),
+            //cmocka_unit_test_setup(test_rdb_tcp_timeout, setupTest), /* too long to run */
```
tezc, line 657: commented out deliberately?

tezc: ok I saw the above comment

src/ext/respToRedisLoader.c
```diff
@@ -324,18 +332,35 @@ _LIBRDB_API void RDBX_setPipelineDepth(RdbxRespToRedisLoader *r2r, int depth) {
...
                                                             RdbxToResp *rdbToResp,
                                                             RdbxRedisAuth *auth,
                                                             int fd) {
-    RdbxRespToRedisLoader *ctx;
-    if ((ctx = RDB_alloc(p, sizeof(RdbxRespToRedisLoader))) == NULL) {
-        RDB_reportError(p, (RdbRes) RDBX_ERR_RESP_FAILED_ALLOC,
-                        "Failed to allocate struct RdbxRespToRedisLoader");
+    /* Ensure the socket is in blocking mode */
+    int flags = fcntl(fd, F_GETFL, 0);
+    if (flags == -1 || fcntl(fd, F_SETFL, flags & ~O_NONBLOCK) == -1) {
```
tezc, line 342: RDBX_createRespToRedisFd() is a public API right? It accepts an fd, then changes some of its socket configs. Now, user has a non-blocking socket with a different timeout but the user is not aware of. I guess it is better to revert when we are done with the fd or at least document it?

moticless: revert. thanks.

tezc (approved):

PR #72, opened 2025-03-30
Fix: Remove from Makefile tests redundant -L /usr/lib

moticless:
Might caused tests to consume installed librdb library rather than compiled one


moticless:
Merge the fix via another PR

PR #73, opened 2025-04-26
Define LIBRDB_VERSION using a portable sed command.

collinfunk:
The 'grep -P' option is a GNU extension for Perl regular expressions. The default grep on BSD's and MacOS do not support this. This patch uses a simple 'sed' BRE that is portable anywhere.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ else
 	PKGCONFIGDIR = $(LIBDIR)/pkgconfig
 endif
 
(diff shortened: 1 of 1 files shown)


moticless (approved):

collinfunk:
Thanks! I can build and use `rdb-cli` on MacOS now.

collinfunk:
Although, I cannot install it to my local user due using:
```
$ make PREFIX=~/.local install
```
since the variable definition in the Makefile has a typo. I have opened https://github.com/redis/librdb/pull/74 to address this.

moticless:
@collinfunk, , your input on macOS is really helpful—thank you!

PR #75, opened 2025-05-14
Fix installation on MacOS without GNU Coreutils.

collinfunk:
The '-r' option to 'ln' is a GNU extension that is not supported by 'ln' on MacOS. Therefore, we can limit PREFIX to canonical paths to avoid its use.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,11 @@
 
 PREFIX ?= /usr/local
 
+# Avoid the need for 'ln -r'.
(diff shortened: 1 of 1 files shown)


collinfunk:
I think that this limitation should be quickly understood since Autoconf places similar restrictions on `--prefix`:

```
$ ./configure --prefix=./test
configure: error: expected an absolute directory name for --prefix: ./test
```

moticless (approved):
LGTM. Thanks.

PR #103, opened 2026-06-23
Add 'stat' command and per-key memory estimation to rdb-cli

moticless:
## Summary

Adds memory-oriented capabilities to `rdb-cli`, surfaced through its generated output:

1. A new **`stat`** command that prints a human-formatted, estimated memory-statistics report.
2. New per-key **memory specifiers for `print`** (`%n` encoding, `%z` estimated bytes, `%g` largest element), plus optional printf-style field widths.

Both share a single key-size estimator, so the per-key `%z` number and the `stat` aggregates are computed the same way.

## `stat` — generated output

```
$ rdb-cli dump.rdb stat
Statistics (Memory is estimated):
                                      [     Expiry     ] [     Memory    ] [          Memory per key         ] [      Items per key       ]
  type             keys         items  volatile  expired        mem   mem%      avg      p90      p99      max    avg    p90    p99     max
  hash          1560278      81456780         0        0       6.8G  74.8%     4.6K    16.0K    64.0K    38.2M     52    128    512  858.8K
  set           7651482      27952535         3        3       1.5G  16.9%     216B     512B     2.0K    61.2M      3      8     32    1.4M
  list          6017862       7906342         0        0     719.9M   7.7%     125B     256B     512B     8.2K      1      2      4    4.1K
  string         106018             -         2        2      63.5M   0.7%     628B     1.5K     4.0K   512.0K      -      -      -       -
  stream              1           948         0        0      22.3K   0.0%    22.3K    22.3K    22.3K    22.3K    948    948    948     948
  TOTAL        15335641     117316605         5        5       9.1G

  Volatile keys hold 2.0K (0.0% of memory)
  Estimated keyspace tables (dict overhead): 586.5M
  Estimated dataset memory: 9.7G (objects + tables; excludes server/client/frag)

Top 10 keys by memory:
        memory type        db        items   avg item      ttl  key
         61.2M set          0      1369515        46B     none  App:AccessGroup:001
         57.4M set          0      1290404        46B     none  App:EntityIndex:All
         38.2M hash         0       858814        46B     none  App:EntityMetadata:002
  ...
```
Extending `rdb-cli` Usage:
```
OPTIONS:
	-n, --now <UNIX-TIME-SEC>     Reference time for expiry evaluation (Default: now)
...
FORMAT_OPTIONS ('stat'):
	-t, --top <N>                 Show the top N keys by estimated memory (Default: 10)
	-h, --histogram <0|1|2>       Append per-type distribution histograms: 0=items/key, 1=memory/key, 2=both

```

## `stat` — value distribution (`--histogram`)

The per-key `avg`/`p90`/`p99`/`max` columns above are derived from per-type log-scale histograms. Pass `--histogram <0|1|2>` (`0`=items/key, `1`=memory/key, `2`=both) to append the full non-empty-bucket distribution, each bucket labelled by its lower bound:

```
$ rdb-cli dump.rdb stat --histogram 0

Histograms (non-empty buckets; min = bucket lower bound, keys = key count, share = % of type, cum% = cumulative share):

  hash / items per key
           min           keys   share     cum%
             8 :        93617    6.0%     6.0%
            16 :       187233   12.0%    18.0%
            32 :       530495   34.0%    52.0%
            48 :       343261   22.0%    74.0%
            64 :       202836   13.0%    87.0%
           128 :        93617    6.0%    93.0%
           256 :        62411    4.0%    97.0%
           512 :        31206    2.0%    99.0%
          2.0K :        14043    0.9%    99.9%
        524.3K :         1560    0.1%   100.0%
```

## `print` — extended per-key memory output

```
$ rdb-cli dump.rdb print --key "db=%d,key=%k,type=%t,enc=%n,bytes=%z,largest=%g,count=%i"
db=0,key=mylist1,type=list,enc=quicklist,bytes=66,largest=2,count=1
db=0,key=lzf_compressed,type=string,enc=string,bytes=180,largest=118,count=0
...
```

New specifiers for `--key`:

| Specifier | Meaning |
|-----------|---------|
| `%n` | Encoding (e.g. `quicklist`, `listpack`, `intset`, `skiplist`) |
| `%z` | Estimated in-memory size in bytes |
| `%g` | Size in bytes of the largest element |

Specifiers also accept an optional width, e.g. `%-20k`, `%10z`.

## Implementation notes

- New `handlersCommon` module holds the shared key-size accumulator, the in-memory size estimator, and file/lifecycle plumbing; both the `print` and `stat` handlers compose `RdbxHandlersBase` and register their own data callbacks over it.
- The size estimate uses rough per-encoding overhead constants in the spirit of Redis `MEMORY USAGE` — calibrated so large keys land within ~25% of live `MEMORY USAGE`. It is explicitly an **estimate**.
- Shared JSON/string escaping moved to `extCommon` (`rdbxOutputPlainEscaping`).
- New public API: `RDBX_createHandlersToStat(p, topN, nowSecs, outFilename)`.
- Tests added in `test/test_rdb_to_mem_stat.c`.


src/ext/handlersFilter.c
```diff
@@ -451,17 +457,17 @@ static RdbxFilter *createHandlersFilterCommon(RdbParser *p,
...
 }
 
 _LIBRDB_API RdbxFilter *RDBX_createHandlersFilterDbNum(RdbParser *p, int dbnum, uint32_t exclude) {
-    return createHandlersFilterCommon(p, NULL, NULL, &dbnum, 0, exclude);
+    return createHandlersFilterCommon(p, NULL, NULL, &dbnum, 0, 0, exclude);
 }
 
-_LIBRDB_API RdbxFilter *RDBX_createHandlersFilterExpired(RdbParser *p, uint32_t exclude) {
-    return createHandlersFilterCommon(p, NULL, NULL, NULL, 1, exclude);
+_LIBRDB_API RdbxFilter *RDBX_createHandlersFilterExpired(RdbParser *p, long long nowSecs, uint32_t exclude) {
```
tezc, line 471: isn't that a breaking change? we don't use this in any other project?

src/ext/handlersCommon.c
```diff
@@ -0,0 +1,207 @@
...
+static int estimIsIntegerStr(const char *s, size_t len) {
+    if (len == 0 || len > 20) return 0;
+    size_t i = (s[0] == '-') ? 1 : 0;
+    if (i == len) return 0;
+    if (s[i] == '0' && len - i > 1) return 0; /* no leading zeros */
+    for (; i < len; i++) if (s[i] < '0' || s[i] > '9') return 0;
+    return 1;
+}
+
+static uint64_t estimValueBytes(int opcode, uint64_t n, uint64_t s,
```
tezc, line 91: I suppose for certain cases, the calculation will be dramatically of. Not sure if we can improve but AI points some cases about stream consumer groups, sparse arrays and intset not being 4 byte per element. Maybe you can evaluate if these cases can be improved. Just fyi.

src/ext/handlersToStat.c
```diff
@@ -0,0 +1,514 @@
...
+     * fails. Order by db, then key name. */
+    if (ea->dbnum != eb->dbnum)
+        return (ea->dbnum > eb->dbnum) - (ea->dbnum < eb->dbnum);
+    return strcmp(ea->key, eb->key);
+}
+
+/* Human-readable byte size. Rotating buffers so several calls in one fprintf
+ * (e.g. memory + avg) don't alias. */
+static const char *human(uint64_t b) {
+    static char ring[8][24]; static int r = 0;
```
tezc, line 207: Static buffers might be problematic if an app has two instances of library and call the same function at the same time. Though, this is just statsw, expectation is that people will call from command line. So, I suppose this is okay. Just fyi, if you notice something else about this issue.

PR #107, opened 2026-08-12
Support RDB v15 - Hash Templates

moticless:
Accept RDB format version 15 (Redis 8.10) and parse the new hash templates,
where a hash's field names are stored once per template rather than per key:

- **Referenced out-of-band** - a top-level `RDB_OPCODE_HASH_TEMPLATE` section, referenced by
  `RDB_TYPE_HASH_TMPL_LP_REF` (30) / `_ARRAY_REF` (32). This is what RDB save writes.
- **Self-contained** - `RDB_TYPE_HASH_TMPL_LP` (29) / `_ARRAY` (31), carrying the field names
  inline (`[fields_fmt][field names][values]`, both `FIELDS_LP` and `FIELDS_RAW`). Redis writes
  these only via `DUMP`, but its loader accepts them from an RDB file too, so librdb does the same.

Either form surfaces through the existing hash callbacks (`handleHashField` at DATA level,
`handleHashPlain` at STRUCT level) - no new callbacks.

### support-restore

A REF-encoded hash points at the out-of-band template section, so its bytes can't be RESTORE'd
as-is:

- **v15+ target** - the parser inlines the field names into a self-contained payload and emits
  `RESTORE`, preserving the source encoding (`_LP_REF` -> `_LP` with the values listpack copied
  verbatim, `_ARRAY_REF` -> `_ARRAY`) so the destination keeps the same memory layout.
- **older target** - falls back to plain `HSET`, portable to any version.

### API

Additive only: new error codes `RDB_ERR_HASH_TMPL_INVLD` and `RDB_ERR_HASH_TMPL_UNKNOWN_ID`.

### Hardening

Template ids and declared field counts come straight off the wire and no longer size allocations,
so a hand-crafted RDB can't crash the parser. Corrupt-input fixtures cover zero/unsorted/duplicate
/unknown template ids, huge ids and field counts, and bad `fields_fmt` in the self-contained forms.

### Tests

- **RESP level** (`test_rdb_to_resp.c`) - pins which command carries the data per target version:
  `RESTORE` for v15+, `HSET` fallback below. Target-driven, so it runs everywhere.
- **Live server** (`test_rdb_to_redis.c`) - `test_rdb_to_redis_hash_template` lets Redis 8.10 build
  both template encodings, saves an RDB and replays it in both modes, asserting the encoding
  survives `RESTORE`. `test_rdb_to_redis_v15_hash_template` replays ready-made v15 fixtures, so the
  `HSET` path is covered on any target version.
- **JSON level** (`test_rdb_to_json.c`) - value edges (integer, empty, LZF) and the corrupt-input
  fixtures above.

### CI

The matrix now covers every RDB version the parser claims to support: added **8.6** (v13),
**8.8** (v14) and **8.10** (v15, on ubuntu and macOS), and dropped a duplicated 7.4 job.
Also `fail-fast: false` so one old-version break no longer cancels the newer jobs, and the
test-modules version gate no longer misreads `8.10` as 8.1 (`bc` -> integer major/minor compare).

---

> [!NOTE]
> **Medium Risk**
> Large new parser paths for untrusted RDB input; mitigated by bounded ids, streaming field reads, and extensive corrupt-fixture tests. RESTORE rewriting affects migration to v15+ only when support-restore is enabled.
> 
> **Overview**
> Adds **RDB format version 15** (Redis 8.10) with **hash templates**: field names stored once per template instead of per key. New types and `RDB_OPCODE_HASH_TEMPLATE` are wired through the parser; decoded hashes still use existing **`handleHashField` / `handleHashPlain`** callbacks.
> 
> **Referenced** hashes (`_LP_REF` / `_ARRAY_REF`) load templates from the RDB template section. **Self-contained** forms (`_LP` / `_ARRAY`) carry inline field names (DUMP-style), matching Redis loader behavior.
> 
> With **`--support-restore`**, REF payloads are rewritten into self-contained RESTORE blobs on v15+ targets (preserving listpack vs array encoding); older targets get **HSET**. New errors: `RDB_ERR_HASH_TMPL_INVLD`, `RDB_ERR_HASH_TMPL_UNKNOWN_ID`. Untrusted template ids and field counts no longer drive allocation sizing.
> 
> CI expands Redis matrix (8.6–8.10), sets **`fail-fast: false`**, fixes **8.10** vs **8.1** in the test-modules gate, and switches cmocka clone URL.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 84b2327da50a191d3b7aba4679dab7e26e868274. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/lib/defines.h
```diff
@@ -38,10 +38,22 @@
 #define RDB_TYPE_STREAM_LISTPACKS_4 26        /* Stream with IDMP support */
 #define RDB_TYPE_STREAM_LISTPACKS_5 27        /* Stream with XNACK / NACK zone (v14) */
 #define RDB_TYPE_ARRAY              28        /* Sparse array (v14) */
-#define RDB_TYPE_MAX                29
+#define RDB_TYPE_HASH_TMPL_LP       29        /* Hinted hash (v15): self-contained (DUMP): [fields_fmt][fields][lp_blob] */
+#define RDB_TYPE_HASH_TMPL_LP_REF   30        /* Hinted hash (v15): with template ref (RDB save): [tid][lp_blob] */
+#define RDB_TYPE_HASH_TMPL_ARRAY    31        /* Hinted hash (v15): self-contained (DUMP): [fields_fmt][fields][v0]...[vN-1] */
+#define RDB_TYPE_HASH_TMPL_ARRAY_REF 32       /* Hinted hash (v15): with template ref (RDB save): [tid][v0]...[vN-1] */
+#define RDB_TYPE_GCRA               33        /* GCRA object (Redis Ent ENABLE_GCRA only; reserved, not parsed) */
```
tezc: why are we adding this? we don't even need to reserve it because its value will change when we add a new type, no?

src/lib/defines.h
```diff
@@ -38,10 +38,22 @@
...
+#define RDB_TYPE_MAX                34
 
 
+/* Highest hinted-hash-template id the registry accepts (v15). The saver assigns
+ * dense ids from 0, one per distinct field-set, so this is far above any real
+ * save. It bounds the id-indexed registry against a crafted sparse id. Note
+ * Redis's loader keys templates by id in a dict and so accepts any id. */
+#define RDB_HASH_TMPL_MAX_ID  (1<<21) /* 2MB entries */
```
tezc: why are having a cap limit? to avoid maintaining hashmap in the lib? btw, 1 million template is unexpected but it could still be possible. If we have to set a cap, maybe we should make it higher.

src/lib/parser.c
```diff
@@ -92,6 +92,11 @@ struct ParsingElementInfo peInfo[PE_MAX] = {
         [PE_STREAM_LP]        = {elementStreamLP, "elementStreamLP", "Parsing stream Listpack"},
         /* array (sparse-array, v14+) */
         [PE_ARRAY]            = {elementArray, "elementArray", "Parsing sparse array"},
+        /* hinted hash templates (v15+) */
+        [PE_HASH_TEMPLATE]      = {elementHashTemplate, "elementHashTemplate", "Parsing hash template record"},
+        [PE_HASH_TMPL_LP_REF]   = {elementHashTmplLpRef, "elementHashTmplLpRef", "Parsing template hash (listpack ref)"},
+        [PE_HASH_TMPL_ARRAY_REF]= {elementHashTmplArrayRef, "elementHashTmplArrayRef", "Parsing template hash (array ref)"},
+        [PE_HASH_TMPL_SELF]     = {elementHashTmplSelfContained, "elementHashTmplSelfContained", "Parsing template hash (self-contained)"},
```
tezc, line 99: does that mean we'll be able to parse DUMP format if RDB contains one? ROF may store in dump format if I'm not mistaken.

src/lib/parser.c
```diff
@@ -2858,6 +2925,540 @@ RdbStatus elementArray(RdbParser *p) {
...
+/* Emit the (field,value) pairs of a values listpack whose first entry is the
+ * template id and remaining entries are the values in field order. Caller has
+ * already validated that 'lp' holds t->fieldCount+1 entries. The values are
+ * passed to the callbacks embedded (referenced) within the listpack. */
+static RdbStatus hashTemplateEmitLpValues(RdbParser *p, unsigned char *lp, HashTemplate *t) {
+    unsigned char *iter = lpFirst(lp); /* skip the leading template-id entry */
+    unsigned int slen;
+    long long v;
+    for (uint64_t i = 0; i < t->fieldCount; ++i) {
+        iter = lpNext(lp, iter);
```
tezc: maybe we should make sure that lpNext() will return t->fieldCount entries?

src/lib/parserRaw.c
```diff
@@ -1467,6 +1469,222 @@ RdbStatus elementRawArray(RdbParser *p) {
     }
 }
 
+/*** hinted hash templates (v15): raw -> self-contained RESTORE ***/
```
tezc: we didn't use "hinted" word in redis code base, just hash template (and product chose "compact hashes"), maybe we should trim hinted word from the comments.

src/lib/parserRaw.c
```diff
@@ -1467,6 +1469,222 @@ RdbStatus elementRawArray(RdbParser *p) {
...
+                RDB_reportError(p, RDB_ERR_HASH_TMPL_INVLD,
+                                "elementRawHashTmplSelfContained(): field blob integrity check failed");
+                return RDB_STATUS_ERROR;
+            }
+
+            IF_NOT_OK_RETURN(cbHandleBegin(p, DATA_SIZE_UNKNOWN_AHEAD));
+            IF_NOT_OK_RETURN(aggUpdateWritten(p, hdrLen));  /* fields_fmt byte */
+            IF_NOT_OK_RETURN(aggWriteRawString(p, blob, blobBulk->len));
+            h->numFields = lpLength(blob);
+            h->visitingField = 0;
```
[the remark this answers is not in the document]
tezc: this seems correct, please take a look

src/lib/defines.h
moticless: removed.

src/lib/parser.c
moticless: Not out of the box. There's no entry point for it. but the hard part is already there, and the v15 work actually makes it easier for template hashes (DUMP/RESTORE blobs are [type byte][value in RDB encoding][2-byte RDB version][8-byte CRC64])

src/lib/defines.h
moticless: Initially I did bring dict for this, but it felt too much, too complex, and slower than the common case deserves. The saver assigns ids densely from 0, so a flat array indexed by id is O(1) with trivial code; a dict only wins for sparse ids, which never occur in a legit save. I would cliam, we should do the same in Redis as well, if it is not too late. Icreased to 8M entries.

tezc: I agree this number should be a lot smaller but bringing this limit into Redis (to runtime) would also require some extra code. e.g. to avoid template creation or to convert key back to regular hash when this limit is reached. Anyway, hope this new limit will be enough.

CHANGELOG.md
```diff
@@ -1,3 +1,26 @@
...
+  - `--support-restore`: a REF-encoded hash points at the out-of-band template
+    section, so its bytes can't be RESTORE'd as-is. Against a v15+ target the
+    parser inlines the field names into a self-contained payload and emits
+    `RESTORE`, preserving the source encoding (`_LP_REF` → `_LP` with the values
+    listpack copied verbatim, `_ARRAY_REF` → `_ARRAY`) so the destination keeps
+    the same memory layout. Against an older target it falls back to `HSET`.
+  - New error codes `RDB_ERR_HASH_TMPL_INVLD`, `RDB_ERR_HASH_TMPL_UNKNOWN_ID`.
+    Template ids and declared field counts read off the wire no longer size
+    allocations, so a hand-crafted RDB can't crash the parser.
+  - `RDB_TYPE_GCRA` (33) is reserved only (Redis Enterprise `ENABLE_GCRA`
```
tezc: we can also remove GCRA references here

src/lib/parser.c
```diff
@@ -2858,6 +2925,555 @@ RdbStatus elementArray(RdbParser *p) {
...
+
+    if (id > RDB_HASH_TMPL_MAX_ID) {
+        RDB_reportError(p, RDB_ERR_HASH_TMPL_INVLD,
+                        "Hash template id %llu exceeds the supported maximum (%llu)",
+                        (unsigned long long) id, (unsigned long long) RDB_HASH_TMPL_MAX_ID);
+        return RDB_STATUS_ERROR;
+    }
+
+    uint64_t newCap = p->hashTmpls.cap ? p->hashTmpls.cap : 16;
+    while (newCap <= id) newCap *= 2;
```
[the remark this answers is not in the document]
tezc: this seems correct?

src/lib/parserRaw.c
```diff
@@ -1467,6 +1469,225 @@ RdbStatus elementRawArray(RdbParser *p) {
...
+
+    switch (ctx->state) {
+        case ST_FMT: {
+            uint64_t fmt, count = 0;
+            int hdrLen = 0, cntLen = 0;
+
+            /* Reserve room for both length words up front so the two reads stay
+             * contiguous. Status safely ignored: first bulk surely has room, and
+             * a maxRawSize failure is re-caught by the next checked call. */
+            aggMakeRoom(p, 18); /* worst case 9 bytes each for fmt + count */
```
tezc, line 1581: need to check return value? btw in the code base, there are some more cases we don't check it as well. I'm not sure if we need to check in all places.

moticless: I wrote in the comment, “Status safely ignored: first bulk surely has room,” because a static 32 KB buffer is used to serialize RESTORE/RAW commands. Therefore, when starting to serialize a new key, the initial write is guaranteed to fit in the buffer, so a potential malloc failure cannot occur at that point.

tezc (approved):

